PR #465, opened 2014-02-05
Add `ActivatorQuerySet` and `ActivatorModel` tests

chrisforrette:
I've been having a problem with not being able to chain `active` and `inactive` calls using the `ActivatorModel`, so I've added `ActivatorQuerySet` to allow for this, and I've added tests for `ActivatorModel`. All tests passing and flake8 runs clean.

I would also love to help getting the `activate_date` and `deactivate_date` stuff working. If someone has something in particular envisioned for this, please let me know, otherwise I'm happy to propose something.

Thanks!

### DIFF
--- a/django_extensions/db/models.py
+++ b/django_extensions/db/models.py
@@ -67,7 +67,7 @@ class ActivatorModelManager(models.Manager):
         return ActivatorQuerySet(model=self.model, using=self._db)
 
     def active(self):
-        """ Returns active instances of ActivatorModel: SomeModel.objects.active(), 
+        """ Returns active instances of ActivatorModel: SomeModel.objects.active(),
         proxy to ActivatorQuerySet.active """
         return self.get_query_set().active()
 
--- a/django_extensions/tests/test_models.py
+++ b/django_extensions/tests/test_models.py
@@ -1,5 +1,3 @@
-import datetime
-
 from django.test import TestCase
 
 from django_extensions.db.models import ActivatorModel


django_extensions/db/models.py
```diff
@@ -41,17 +41,40 @@ class Meta:
...
     Manager to return instances of ActivatorModel: SomeModel.objects.active() / .inactive()
     """
+    def get_query_set(self):
+        """ Proxy to `get_queryset`, drop this when Django < 1.6 is no longer supported """
+        return self.get_queryset()
+
+    def get_queryset(self):
+        """ Use ActivatorQuerySet for all results """
+        return ActivatorQuerySet(model=self.model, using=self._db)
+
```
blueyed: @chrisforrette 
I think these get_queryset methods are problematic, because Django might use them automatically.

I am not sure, but this might get changed to be a private/local function?

Otherwise: can this be made to call `super` / parent classes (easily)?
Maybe similar to QueryManagerMixin from https://github.com/carljm/django-model-utils/blob/master/model_utils/managers.py#L198, where it also extends `__init__`.

With Django 1.7 the default is:

    def get_queryset(self):
        return self._queryset_class(self.model, using=self._db, hints=self._hints)

I had problems because this `get_queryset` method overwrote the queryset from `objects = BaseManager.from_queryset(BaseQuerySet)()` being used.

I am now using `objects = MyManager.from_queryset(MyQuerySet)()`, where `MyQuerySet` inherits from ActivatorQuerySet.

PR #595, opened 2015-01-09
Updating tox.ini to latest Django versions, updated docs for JSONField

jaddison:
Most importantly, the 1.7 references were to the 'b3' prelease.


trbs:
Thanks !

PR #596, opened 2015-01-12
Fix check for missing import_helper module in dumpscript (Python 3.3+).

safehammad:
Since Python 3.3 the ImportError message has been changed from
"No module named foo" to "No module named 'foo'" (note the quotes).
This broke the check for a missing import_helper module when running
a script generated by the dumpscript management command.


trbs:
Thanks !

PR #598, opened 2015-01-19
shell_plus: print notice about auto-used shell

blueyed:
This is useful.

Maybe the default verbosity should get changed though, so that it would only get displayed when requesting verbose output (verbosity=2)?


trbs:
Thanks again ! :)

I have changed the level to verbosity 2

blueyed:
:+1:

PR #599, opened 2015-01-19
shell_plus: look at $EDITOR to detect vi_mode

blueyed:
If you are using `EDITOR=vim`, `vi_mode` should default to `True`.


trbs:
Thanks !

Issue #454, opened 2014-01-11
Test Runner Issue


camilonova:
Somehow django 1.6 does not run all the tests:

https://travis-ci.org/django-extensions/django-extensions/jobs/16780166#L132

It was running only 6, meanwhile other django versions where running 35 tests:

https://travis-ci.org/django-extensions/django-extensions/jobs/16780167#L173

So I have made this change https://github.com/django-extensions/django-extensions/commit/5285fd664c7ff4c0fee98aef03de8d1d91b22c1d

It works for django 1.6 but somehow does not work for django 1.5 and 1.4:

https://travis-ci.org/django-extensions/django-extensions/builds/16789212

Maybe the best solution would be to change the way we use the test runner, but since I don't know much about it I'll leave it here.

camilonova:
Maybe one year later? ;)

trbs:
Hehe :) It's clearly not been on my priority list... sorry about that.

Right now Travis seems to indicate that we run 57, 57, 40 and 40 tests for Django 1.7, 1.6, 1,5 and 1,4 respectively. Since 1.5 and 1.4 are very old now and 1.5 is unsupported I don't think we have to care too much about those.

However the tests on 1.8 are broken so we might need to take a good look at the testsuite anyways.

camilonova:
Yay!

trbs:
@camilonova what are your thoughts on replacing our current test suite with something based on py.test like suggested by https://github.com/benjaminabel ?

Think that might solve this issue as well right ?

camilonova:
Yes it is.

@benjaminabel what do you think?

benjaminabel:
@camilonova  `py.test django` find 59 tests for all django versions, so I think it could close the issue. However this need a little more work because all tests don't pass for all django versions.

benjaminabel:
@camilonova @trbs I'm moving `tests` directory to root directory, and renaming each test to `test_....py`.
It's possible to run tests as functions in pytest, however we could use django.test.TestCase class, what do you think?

camilonova:
@benjaminabel Sounds great for me. thanks

benjaminabel:
@camilonova @trbs the pr is ready for a review.

PR #611, opened 2015-02-02
Use TOXENV variables in `travis.yml`

benjaminabel:


trbs:
Thanks !

benjaminabel:
@trbs Would you like to use django py.test in your test suite?

trbs:
@benjaminabel maybe we can take this to https://github.com/django-extensions/django-extensions/issues/454 ?

There we have a small discussion on the way we currently do testing. I think moving to py.test and their best practices might be very good move to improve our test(runner).

benjaminabel:
@trbs so you seem ok.  Moving to #454

trbs:
Great thanks

PR #612, opened 2015-02-02
Fix 1_7.W001: Django 1.7 for the MIDDLEWARE_CLASSES

benjaminabel:
See <https://docs.djangoproject.com/en/1.7/ref/checks/#backwards-compatibility>


trbs:
Thanks !

Issue #353, opened 2013-07-30
Command print_user_for_session only works with DB backend


glarrain:
It's fine but it should say so

trbs:
please submit a pull request to update the documentation on this.

PR #613, opened 2015-02-03
docs: add note about `print_user_for_session`

glarrain:
As suggested in https://github.com/django-extensions/django-extensions/issues/353


trbs:
Great thanks !

PR #616, opened 2015-02-11
Add support for `extra_files` to runserver_plus

vikynandha:
Werkzeug [provides](http://werkzeug.pocoo.org/docs/0.10/serving/) `extra_files` options to specify files whose changes, in addition to the default ones, should auto-reload the server. However, `runserver_plus` of `django_extensions` currently doesn't support this option.

This pull request adds support for the option.


trbs:
Thanks !

I changed it slightly so that --extra-file can be specified multiple times as command line arguments.
Could you please test and make sure it works as expected from the git version ?

vikynandha:
Yep, works fine. Thanks!

PR #617, opened 2015-02-11
Fixing shell_plus problem when auto-loading modules with empty `__module__` property.

ewh:
Only add the a module to the `load_models` dict if `mod.__module__` is not empty. The `__module__` property can be empty for dynamically generated modules.


trbs:
Thanks !

PR #623, opened 2015-02-18
Add load commandclass test and compat file

benjaminabel:
I added a test to check if management command classes can be loaded in every environment.

To fix import errors revealed by this new test I added a `compat.py` file to gather compatibility code between python versions(and maybe also between django versions ). With this it could be simpler to remove unused compatibility code in the future.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -3,7 +3,6 @@ from django.conf import settings
 
 def pytest_configure():
     import sys
-    import tempfile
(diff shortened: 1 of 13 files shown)


trbs:
Thanks you rocking it man :)

benjaminabel:
It seems that coveralls liked the for loop;)

PR #632, opened 2015-03-02
Moved content of future_1_5 in the new compat file

benjaminabel:
I think this is cleaner and future proof as django 1.4 will be deprecated in about 6 mounths.


trbs:
Thanks !

Issue #631, opened 2015-03-01
shell_plus --notebook gives bad configuration error


kutenai:
I am getting the following error when trying to use shell_plus --notebook:

[C 04:28:24.002 NotebookApp] Bad config encountered during initialization:
[C 04:28:24.003 NotebookApp] Unrecognized flag: '--ext'

This is my pip environment:
backports.ssl-match-hostname==3.4.0.2
certifi==14.5.14
Django==1.7.4
django-account==0.1.14
django-appconf==1.0.1
django-bootstrap-form==3.1
django-debug-toolbar==1.2.2
django-extensions==1.5.0
django-jsonfield==0.9.13
django-registration==1.0
django-suit==0.2.12
django-urlauth==0.1.8
django-user-accounts==1.0.1
djangorestframework==3.0.5
docutils==0.12
eventlog==0.10.0
gnureadline==6.3.3
ipython==3.0.0
Jinja2==2.7.3
jsonschema==2.4.0
MarkupSafe==0.23
metron==1.3.5
mistune==0.5
MySQL-python==1.2.5
nose==1.3.4
numpydoc==0.5
pinax-theme-bootstrap==5.7.1
psycopg2==2.6
ptyprocess==0.4
Pygments==2.0.2
pytz==2014.10
PyYAML==3.11
pyzmq==14.5.0
requests==2.5.3
sh==1.11
simplejson==3.6.5
six==1.9.0
Sphinx==1.2.3
sqlparse==0.1.14
terminado==0.5
tornado==4.1

davecoates:
This is something to do with ipython 3.0.0 - I upgraded to this and got the same error. After downgrading to 2.4.0 at it works again.

trbs:
Skimming over the backwards-incompatible-changes list of the IPython 3.0 release I could not find the culprit. But assuming either the '--ext' option has been renamed or loading extensions has been changed all together I am assuming this should be relatively easy to fix.

pebreo:
This cause seems to be diagnosed here: https://github.com/ipython/ipython/issues/7913

trbs:
Hopefully it is still possible to tell the notebook to load those extension for Python notebooks some other way.

```IPython 3, as part of the language-agnostic kernel starting changes, no longer relays the notebook's CLI arguments to the kernel.```

Otherwise it might be the end of ```shell_plus --notebook```

takluyver:
[...]

PR #634, opened 2015-03-04
use kernelspec to pass IPython kernel args

minrk:
in IPython >= 3

This is not necessarily a proposal for merge, but an illustration for discussion of one of the possibilities in #631

- kernelspec is always created as 'django', so multiple django apps will all write the same file. Alternately, a kernelspec could be created in a distinct location for each django app.
- separates notebook and IPython args
- specifies `--notebook-dir` explicitly, in case a user has a default notebook path in config
- behavior should be unchanged for IPython < 3

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -121,11 +121,21 @@ class Command(NoArgsCommand):
                 except ImportError:
                     return traceback.format_exc()
 
+            def install_kernel_spec(app, ipython_arguments):
+                """install an IPython >= 3.0 kernelspec that loads django extensions"""
+                ksm = app.kernel_spec_manager
+                ks = ksm.get_kernel_spec('python')
+                ks.argv.extend(ipython_arguments)
+                ks.display_name = 'Django Admin'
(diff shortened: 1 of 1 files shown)


django_extensions/management/commands/shell_plus.py
```diff
@@ -121,11 +121,21 @@ def get_notebook():
                 except ImportError:
                     return traceback.format_exc()
 
+            def install_kernel_spec(app, ipython_arguments):
+                """install an IPython >= 3.0 kernelspec that loads django extensions"""
+                ksm = app.kernel_spec_manager
+                ks = ksm.get_kernel_spec('python')
+                ks.argv.extend(ipython_arguments)
+                ks.display_name = 'Django Admin'
```
minrk: here, the display name is "Django Admin", but it can be whatever you want. It can include the project name, Python version etc. (it should be kept relatively short). This is what users see when picking the kernel in a new-notebook dropdown.

minrk: it would also be possible to use the KernelSpec whitelist, to make this the only available kernel.

trbs:
Thanks ! Looks like a great basis for IPython 3 support.

PR #638, opened 2015-03-12
Added try/catch for importing json/simplejson for Django 1.7

NSLockhart:
This has been done in the code base in other places and possibly overlooked in this file.


trbs:
Yeah i think so, thanks for the PR !

PR #644, opened 2015-03-19
Fix runscript with AppConfig apps

### LINKED ISSUE #641: runscript not compatible with apps that use AppConfig in INSTALLED_APPS
Since Django 1.7 (https://docs.djangoproject.com/en/1.7/ref/applications/#configuring-applications), it is possible to use:

```python

INSTALLED_APPS = (
    "my_app.apps.MyAppConfig",
)
```

But runscript fails to load scripts for these apps.
It looks for ``my_app.apps.MyAppConfig.scripts.my_script`` instead of ``my_app.scripts.my_script``.

NotSqrt:
with tests.

Fixes #641

### DIFF
--- a/django_extensions/compat.py
+++ b/django_extensions/compat.py
@@ -3,6 +3,7 @@ from __future__ import unicode_literals
 import sys
 
 import django
+from django.conf import settings
 
 # flake8: noqa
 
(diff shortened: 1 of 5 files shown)


NotSqrt:
Unrelated error for py32-dj16 : ssl error during ``pip install``

trbs:
Fixed the unrelated build error :)

trbs:
Thanks !

PR #646, opened 2015-03-20
Sqldsn

rodo:
Please find here my proposal for a new usefull management command


trbs:
Thanks for the PR.

Hope you don't mind but I added some more options to the command and the ability to show the DSN for all databases.

rodo:
@trbs not at alln thanks for them and the merge, happy to see the PR merged so quick !

PR #649, opened 2015-03-25
Replaces `unreferenced_files` command with `delete_unregistered_media`.

BertrandBordage:
Changes from `unreferenced_files`:
- Deletes files if confirmed.
- Displays a nice progression line instead of thousands of file paths.
- Stores files path and size in a CSV instead of printing paths.
- Uses an optimized SQL query resulting in an about 6 times faster execution.
- Uses almost no RAM (the former command was making
  a list of all files and registered files, which is leads to a MemoryError
  when you have millions of media files).
- Marks easy_thumbnails thumbnails as registered.


trbs:
I would like to provide people with a deprecation timeline for replacing the ```unreferenced_files``` command before removing it. 

Think it's also important that the new command is still able to output the files to the stdout as this is the usual unix way to create more complex operations over that list.

jrd:
Why not adding the new command and only deprecate the old one (or keep both)?

trbs:
Happy to add the command next to the existing one and start deprecation path. But it would need to be rebased and some changes (see above comments) still need to be made.

Also would be good to structure the code such that tests can be added.

For now I will close this PR, if anybody is willing to pick it up then I'm happy to merge an PR.

PR #651, opened 2015-03-30
Fix sync_s3 for python3

Kondou-ger:
@trbs 

I didn't test this with py2 and other regions than eu-central. Kudos if someone could do that for me.


trbs:
Thanks !

PR #656, opened 2015-04-09
Doc `NOTEBOOK_ARGUMENTS` in `shell_plus`.

xychu:
It's 'unfair' that the `NOTEBOOK_ARGUMENTS` did not get mentioned in the docs.

For IPython >= 3.0, the args in `IPYTHON_ARGUMENTS` will not be extended to args in `NOTEBOOK_ARGUMENTS`, so it's better to tell users that the `NOTEBOOK_ARGUMENTS` can be used to hold these options that available via:

    $ ipython notebook -h


trbs:
Thanks !

PR #653, opened 2015-04-03
Django 1.8 support

jpadilla:
- Update tox.ini
- Fix `RemovedInDjango19Warning: "requires_model_validation" is deprecated in favor of "requires_system_checks".`

PR #660, opened 2015-04-17
Django 1.7+ deprecation fixes

Antwan86:
This PR adds fixes for Django 1.7+ replacing all occurrences of the [deprecated meta parameter](https://docs.djangoproject.com/en/dev/releases/1.7/#django-core-management-basecommand) `requires_model_validation` with the new `requires_system_checks`. This removes warning when running associated commands in Django 1.7+.

Note for an unknown reason, these fixes were [marked as done in the changelog](https://github.com/django-extensions/django-extensions/blob/master/CHANGELOG.md#144) but the attributes were still used.

*Also, travis CI now uses Django 1.8 stable code for the tests.*

### DIFF
--- a/django_extensions/management/commands/clean_pyc.py
+++ b/django_extensions/management/commands/clean_pyc.py
@@ -19,7 +19,7 @@ class Command(NoArgsCommand):
     )
     help = "Removes all python bytecode compiled files from the project."
 
-    requires_model_validation = False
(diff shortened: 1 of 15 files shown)


Antwan86:
Actually this is a duplicate of #653. Please merge the PR as this is creating issues when running projects on latest Django stable branch.

Issue #663, opened 2015-04-27
pipchecker error: 'Link' object has no attribute 'startswith'


kevgathuku:
When running `./manage.py pipchecker` against a requirement file with a Github URL. I get this error
```
Traceback (most recent call last):
  File "./manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/home/kevin/.virtualenvs/extensions/lib/python3.4/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/home/kevin/.virtualenvs/extensions/lib/python3.4/site-packages/django/core/management/__init__.py", line 330, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/kevin/.virtualenvs/extensions/lib/python3.4/site-packages/django/core/management/base.py", line 390, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/kevin/.virtualenvs/extensions/lib/python3.4/site-packages/django/core/management/base.py", line 441, in execute
    output = self.handle(*args, **options)
  File "/home/kevin/.virtualenvs/extensions/lib/python3.4/site-packages/django/core/management/base.py", line 657, in handle
    return self.handle_noargs(**options)
  File "/home/kevin/code/django/django-extensions/django_extensions/management/utils.py", line 71, in inner
    ret = func(self, *args, **kwargs)
  File "/home/kevin/code/django/django-extensions/django_extensions/management/commands/pipchecker.py", line 96, in handle_noargs
    self.check_github()
  File "/home/kevin/code/django/django-extensions/django_extensions/management/commands/pipchecker.py", line 190, in check_github
    if req_url.startswith("git") and "github.com/" not in req_url:
AttributeError: 'Link' object has no attribute 'startswith'
```
It seems the `Link` object should be converted to a string first before calling the `startswith` method on it

trbs:
Thanks !

PR #664, opened 2015-04-27
Resolve issue #663

kevgathuku:
Convert `req["url"]` to string before calling `startswith` method on it


trbs:
I committed a slightly different version of your patch to make sure that we don't call ```str()``` when the value is ```None```.

Thanks !

kevgathuku:
Yeah. That's the better way to do it
:+1:

PR #668, opened 2015-05-01
1.8 template changes

matclayton:
Quick patch to make template validation work on 1.8, also ran against 1.7 without issue


trbs:
Thanks !

PR #670, opened 2015-05-04
Update syncdata.py to allow skip remove objects

mzaforas:
I suggest add "skip-remove" option to avoid remove old db objects when you want to maintain it


trbs:
Thanks !

mzaforas:
You're welcome! thanks for your merge

PR #671, opened 2015-05-05
update module_name to model_name

ricefield:
in Django 1.8, the alias module_name was removed, the correct usage is model_name. See: https://docs.djangoproject.com/en/1.8/releases/1.8/#features-removed-in-1-8

### DIFF
--- a/django_extensions/admin/__init__.py
+++ b/django_extensions/admin/__init__.py
@@ -5,6 +5,7 @@ import six
 import operator
 from six.moves import reduce
 
+import django
 from django.http import HttpResponse, HttpResponseNotFound
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


trbs:
Is this fully backwards compatible with all currently (officially) supported Django versions ?

ricefield:
@trbs no, though we could do something similar to what django-guardian does here: https://github.com/lukaszb/django-guardian/pull/287/files#diff-af788baba55f2be5cdad4875792ff6e2R109

trbs:
I would be more then happy with that solution :)

ricefield:
@trbs pull-request updated

trbs:
great thanks !

PR #672, opened 2015-05-07
Force the result of filter into a list

bradmontgomery:
Otherwise, this fails in python 3.4 on line 26 (`apps += template_dirs`) with:

    TypeError: unsupported operand type(s) for +=: 'filter' and 'list'


trbs:
Thanks !

PR #673, opened 2015-05-08
Update sync data: Move dict outside loop to analyze all fixtures files before remove

mzaforas:
When you have fixtures related with the same model split in more than one fixture file is a problem sync them with syncdata because the command create and delete objects on every file iteration, so at the end of the execution you will have the last provided file well sync, but the previous data deleted.

My suggestion to fix the issue is move objects_to_keep dict definition out of the for loop and acum all the objects to keep for all the files on each iteration. Then after the end of the loop try to remove residual objects, based on all the accumulated objects.

### DIFF
--- a/django_extensions/management/commands/syncdata.py
+++ b/django_extensions/management/commands/syncdata.py
@@ -103,6 +103,7 @@ class Command(BaseCommand):
         object_count = 0
         objects_per_fixture = []
         models = set()
+        objects_to_keep = {}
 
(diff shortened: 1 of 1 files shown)


trbs:
Could we add another switch to support something like rsync {{{--delete-during}}} with your patch ?

For very large (or very slow) systems we might want to still give people the option to remove files right away instead of only afterwards.

trbs:
Bump

PR #675, opened 2015-05-17
RandomCharField - Prepopulates random character string.

dpetzold:
Hello,

I added support for a random character field. Its purpose is to create an unique id unrelated to the database id to prevent site walking and sequential attacks on something like a unsubscribe link.  A sample use case would be a link shortening service like bit.ly, http://bit.ly/1AbL4UC.

Non-Breaking changes:
- Factored out query logic to ensure uniqueness to BaseUniqueField.
- The field value factories have been moved to generators.
- Added mock as a test requirement.

Possibly breaking change:
- The slug field check will only be tried 100 times. Previously there was no limit. That was done because the field value factories have been moved to a generators.

Please let me know your comments and feedback. I enjoyed working on this code very much!

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -1,8 +1,10 @@
 """
 Django Extensions additional model fields
 """
+import random
 import re
 import six
+import string
 import warnings
 
 try:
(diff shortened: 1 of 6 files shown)


trbs:
Could you please take a look at why the tests are failing ?
(Sorry I don't have the time right now to dive into this)

dpetzold:
Its from this

    ValidationError: [u'The verbose_name of random char test model lower alpha digits is longer than 39 characters']

I'll fix and try to increase coverage.

trbs:
Ready to merge ?

dpetzold:
Yeah this is it.

dpetzold:
Sorry I have been changing history to keep the commit clean. That is why its rebuilding.

trbs:
Thanks !! :)

PR #676, opened 2015-05-18
get rid of RemovedInDjango19Warning in 'runjobs'

flesser:
Previously, `python manage.py runjobs` gave three RemovedInDjango19Warning errors:

```
[...]/py3env/lib/python3.4/site-packages/django_extensions/management/commands/runjobs.py:51: RemovedInDjango19Warning: django.db.models.get_apps is deprecated.
  for app in models.get_apps():

[...]/py3env/lib/python3.4/importlib/_bootstrap.py:321: RemovedInDjango19Warning: The utilities in django.db.models.loading are deprecated in favor of the new application loading system.
  return f(*args, **kwds)

[...]/py3env/lib/python3.4/site-packages/django/db/models/__init__.py:56: RemovedInDjango19Warning: [a.models_module for a in get_app_configs()] supersedes get_apps().
  return getattr(loading, function_name)(*args, **kwargs)
```

### DIFF
--- a/django_extensions/management/commands/runjobs.py
+++ b/django_extensions/management/commands/runjobs.py
@@ -48,7 +48,7 @@ class Command(LabelCommand):
             except ImportError:
                 pass
 
-        for app in [a.models_module for a in apps.get_app_configs()]:
+        for app in [a.models_module for a in apps.get_app_configs() if a.models_module is not None]:
             if verbosity > 1:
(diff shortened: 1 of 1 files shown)


trbs:
I don't think this is backwards compatible right ?
If not could you please make it so that it will work on all officially supported Django versions ?

trbs:
Implemented this via ```django_extensions.compact.get_apps()``` so that it will work on old and new versions of Django.

PR #679, opened 2015-05-25
Support ptpython>=v0.27.

ryneeverett:
Ptpython was moved to a separate package in v0.27


ryneeverett:
Assuming the CI failures aren't my fault.

trbs:
Could you please make this change backwards compatible ?
So that it does not break for existing users with the old prompt-tool package.

Thanks

ryneeverett:
@trbs It is backwards compatible. Just tested with prompt-toolkit v0.26

trbs:
I think I completely read over the second ```try-except ImportError```.

Thanks !

PR #691, opened 2015-06-14
Random char field followup part deux

dpetzold:
Hey Trbs,

I meant to do what you suggested and use unique instead of db_index. I also noticed if lowercase is set to false only uppercase characters are used now. I do not think that was the best way to have a uppercase only option as now there is no option for mixed case which should be the default as having uppercase only as the default reduces the possible character population by 26 characters a significant amount. I think what is best to have a uppercase option. So I added support for that. @dudanogueira does that sound right?

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -228,6 +228,9 @@ class RandomCharField(UniqueFieldMixin, CharField):
     lowercase
         If set to True, lowercase the alpha characters (default: False)
 
+    uppercase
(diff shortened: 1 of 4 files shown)


trbs:
Looks good :)

dudanogueira:
great!! Not it can serve more people :) Also, the other change I made, for allowing the pre_save method to be triggered even on not adding situation doesn´t seem to work. It works on django admin, but not on the shell. Can´t say exactly why... :(

PR #696, opened 2015-07-01
Support alternative URL conf in `show_urls`

mrben:
This PR adds support for overriding the default settings attribute name for the URL conf. It's useful in circumstances where you need to swap out the root config for something else, e.g. subdomain-specific URLs.

Example `settings.py` snippet:

```python
ROOT_URLCONF = 'mysite.urls'
SUBDOMAIN_URLCONF = 'mysite.subdomain_urls'  # Used for URLs like subdomain.mysite.com/account/edit/
```

Using the URL conf:

```shell
./manage.py show_urls --urlconf=SUBDOMAIN_URLCONF
```

I understand this is a bit of an edge case :smile:. I'm currently working on a site where most of the views/urls are for dynamic subdomains which prompted me to make this amendment.

Regards
Ben.


trbs:
Thanks !

mrben:
:+1:

Issue #483, opened 2014-03-21
Add --pdb and --ipdb options to runserver_plus / all commands?


blueyed:
There is https://github.com/tomchristie/django-pdb, which adds additional options to `runserver` (which does not appear to work in Django 1.7 anymore).

It uses a middleware to start pdb/ipdb in case of errors.

I think this would be a nice addition to django-extensions.

It would be even better, if these options could be added to all management commands somehow globally (requested for django-pdb in https://github.com/tomchristie/django-pdb/issues/19).

tomchristie:
Sure, no objection from me :smile:

trbs:
@blueyed could you setup a Pull Request for this ?

blueyed:
@trbs 
In general, yes - but I'm short on time for things like that currently.  Sorry.

ryneeverett:
I do want this feature, but I'm concerned about the complication this might introduce to `django-extensions` users. That is, in order for the debugging features to work, `django-extensions` users would presumably have to follow [the installation constraints](https://github.com/tomchristie/django-pdb#installation) of `django-pdb`:

- Add to `INSTALLED_APPS` "towards the beginning" in Django>=1.7, otherwise "towards the end".
- Add to `MIDDLEWARE_CLASSES` "after all other middleware".

Questions:

- Can we remove the constraints?
  - If so, shouldn't we do this upstream before trying to integrate?
  - If not, are these constraints acceptable in a popular app that otherwise has no such constraints?
    - If so, then there's no problem and I'm wasting everyone's time.
    - If not, would it be acceptable for these features to be "extra" so the debugging features are disabled if users don't install the optional middleware?

trbs:
Maybe we can leverage the ```django-pdb``` package ?

That way all constraints from ```django-pdb``` will exists if one wants to use it in combination with ```django-extensions```.

For this it would probably be good if we create a custom base ```Command``` class for ```django-extensions``` so that all commands can 'magically' get support for ```--pdb```.

Maybe ```django-pdb``` needs some patches as well so that we can easily hook into it ?

ryneeverett:
I looked into this again and didn't get very far. However `django-pdb` seems work fine without manually adding the middleware so my above concerns may be irrelevant:

*django_pdb/management/commands/runserver.py:55*
```py
        middleware = 'django_pdb.middleware.PdbMiddleware'
        if ((pdb_option or settings.DEBUG)
[...]

PR #703, opened 2015-07-18
Integrate with django-pdb.

ryneeverett:
Per issue #483.

I haven't tested very thoroughly (just with the `django-pdb` testapp) but hoping to get some feedback if I'm on the right track.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -12,7 +12,7 @@ from django.core.management.base import BaseCommand, CommandError
 from django_extensions.management.technical_response import \
     null_technical_500_response
 from django_extensions.management.utils import (
-    RedirectHandler, setup_logger, signalcommand,
+    RedirectHandler, setup_logger, signalcommand, has_ipdb
 )
 
 try:
@@ -64,6 +64,13 @@ class Command(BaseCommand):
                     help='auto-reload whenever the given file changes too (can be specified multiple times)'),
         make_option('--reloader-interval', dest='reloader_interval', action="store", type="int", default=DEFAULT_POLLER_RELOADER_INTERVAL,
                     help='After how many seconds auto-reload should scan for updates in poller-mode [default=%s]' % DEFAULT_POLLER_RELOADER_INTERVAL),
+        make_option('--pdb', action='store_true', dest='pdb', default=False,
+                    help='Drop into pdb shell on at the start of any view.'),
+        make_option('--ipdb', action='store_true', dest='ipdb', default=False,
+                    help='Drop into ipdb shell on at the start of any view.'),
+        make_option(
+            '--pm', action='store_true', dest='pm', default=False,
+            help='Drop into ipdb shell if an exception is raised in a view.'),
     )
     if USE_STATICFILES:
         option_list += (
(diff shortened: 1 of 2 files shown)


ryneeverett:
Ok, as far as I can tell this works just as well as standalone `django-pdb`.

trbs:
@ryneeverett thanks for picking this up !

It looks good, however I would prefer if we wouldn't have django-pdb as a hard requirement.
Could you update the PR so that it passes the flake8 tests and uses django-pdb only if it's available ?

ryneeverett:
The only flake8 errors I'm getting now are for idiomatically importing ipdb and Ipython to see if they're available. I'm assuming it would not be preferable to check `globals()`.

ryneeverett:
Oh wait, that would require it already be imported. Yeah, I'm not aware of any other way to check that.

ryneeverett:
Heh, well I learned something, albeit unuseful.

trbs:
hehe I think easier would be to just add ```# noqa``` at the end of the two import statements :)

ryneeverett:
squashed and green

trbs:
Thanks !

django_extensions/management/commands/runserver_plus.py
```diff
@@ -64,6 +64,13 @@ class Command(BaseCommand):
                     help='auto-reload whenever the given file changes too (can be specified multiple times)'),
         make_option('--reloader-interval', dest='reloader_interval', action="store", type="int", default=DEFAULT_POLLER_RELOADER_INTERVAL,
                     help='After how many seconds auto-reload should scan for updates in poller-mode [default=%s]' % DEFAULT_POLLER_RELOADER_INTERVAL),
+        make_option('--pdb', action='store_true', dest='pdb', default=False,
+                    help='Drop into pdb shell on at the start of any view.'),
```
blueyed: s/on at/at/

django_extensions/management/commands/runserver_plus.py
```diff
@@ -64,6 +64,13 @@ class Command(BaseCommand):
...
+        make_option('--ipdb', action='store_true', dest='ipdb', default=False,
+                    help='Drop into ipdb shell on at the start of any view.'),
```
blueyed: s/on at/at/

django_extensions/management/commands/runserver_plus.py
```diff
@@ -149,9 +156,44 @@ def execute(self, sql, params=()):
...
+        try:
+            from django_pdb.middleware import PdbMiddleware
+
+            pdb_option = options.get('pdb', False)
+            ipdb_option = options.get('ipdb', False)
+            pm = options.get('pm', False)
+
+            # Add pdb middleware if --pdb is specified or if in DEBUG mode
+            middleware = 'django_pdb.middleware.PdbMiddleware'
+            if ((pdb_option or settings.DEBUG) and
```
blueyed: Does this need `or ipdb_option`?

django_extensions/management/commands/runserver_plus.py
ryneeverett: In my defense those typos are in pdb-django too, haha.

django_extensions/management/commands/runserver_plus.py
trbs: I was wondering about that as well

ryneeverett: Yeah, I couldn't figure out what purpose that served in pdb-django but I forgot to try removing it.

blueyed: Without it, --ipdb would not add the middleware?!

ryneeverett: I guess the idea is that you can use pdb-django when `DEBUG=False` if you explicitly pass a flag, but the other ways of entering a debugger are disallowed.

ryneeverett: > Without it, --ipdb would not add the middleware?!

Hmm?

blueyed: Yes, you can explicitly use `--pdb`, but not `--ipdb` currently.

It should be:

    if ((pdb_option or ipdb_option or settings.DEBUG) and

Or move it down and use `settings.DEBUG or PdbMiddleware.always_break``.

ryneeverett: Oh wow, good point.

trbs: committed the first option

PR #710, opened 2015-07-28
Using django-admin instead of manage.py for ipython notebook overwrites PYTHONPATH from environment

haakenlid:
PYTHONPATH set in environmental variable should not be discarded by `shell_plus` management command.

How to reproduce the bug:

    $ export PYTHONPATH="/www/django-project/"
    $ django-admin shell_plus --notebook

In this case `/www/django-project/` does not end up on sys.path. Instead the folder that contains `django-admin` binary will be added.


trbs:
Thanks !

Issue #714, opened 2015-08-01
Set application_name database options by default in shell_plus


rodo:
On production (and other) system it may be important to identify SQL queries origin, there is a mecanism in PostgreSQL called **application_name** I suggests to automatically set this in *shell_plus*, it may be set two in other Database System. It's easy to set this by adding an **OPTIONS** in `settings.DATABASES`, I'll propose a PR to do this

With right log configuration you'll see lines like theses in your logs :

```
[4-1] user=rodo,db=foo,host=10.0.10.1,app=shell_plus LOG:  duration: 50.004 ms  statement: SELECT COUNT(*) FROM bar
[5-1] user=rodo,db=foo,host=10.0.10.1,app=worker1 LOG:  duration: 45.434 ms  statement: SELECT COUNT(*) FROM bar WHERE id = 42
[6-1] user=rodo,db=foo,host=10.0.10.2,app=worker2 LOG:  duration: 42.998 ms  statement: SELECT COUNT(*) FROM bar WHERE id = 37
```

The first two lines just defers by the **application_name** (parameter %a in `log_line_prefix` postgres config entry), as is you can  immediatly know that bad query without a where clause was done by a human using the *shell_plus* and it's not in your code.
 
Ref :  http://www.postgresql.org/docs/9.4/static/libpq-connect.html#LIBPQ-CONNECT-APPLICATION-NAME

trbs:
thanks !

PR #715, opened 2015-08-01
Set application_name database option by default

rodo:
Set application_name database option by default ref #714


trbs:
Cool thanks !

Issue #720, opened 2015-08-12
iPython 4.0 breaks shell_plus notebook


Koed00:
Upgrading to iPython 4.0 will create an error when running ``python manage.py shell_plus --notebook``:

```python
.virtualenvs/django-q/lib/python3.4/site-packages/jupyter_client/kernelspec.py", line 143, in get_kernel_spec
    raise NoSuchKernel(kernel_name)
jupyter_client.kernelspec.NoSuchKernel: 'python'
```

I've tested this with both Python 2.7.9 and 3.4.2.
Downgrading to iPython 3.2.1 returns everything to normal.

multigl:
https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/shell_plus.py#L130

changing 
```python
ks = ksm.get_kernel_spec('python')
``` 
to 
```python
ks = ksm.get_kernel_spec('python2')
```` 
will fix the problem as a quick workaround.

**Comment by aronysidoro** [comment]
I had the same issue with using **iPython 4.0.0**.  This is the error traceback I got:

```
/Users/aaron/Documents/virtualenvs/django18_py2/lib/python2.7/site-packages/IPython/html.py:14: ShimWarning: The `IPython.html` package has been deprecated. You should import from `notebook` instead. `IPython.html.widgets` has moved to `ipywidgets`.
  "`IPython.html.widgets` has moved to `ipywidgets`.", ShimWarning)

/Users/aaron/Documents/virtualenvs/django18_py2/lib/python2.7/site-packages/IPython/frontend.py:21: ShimWarning: The top-level `frontend` package has been deprecated. All its subpackages have been moved to the top `IPython` level.
  "All its subpackages have been moved to the top `IPython` level.", ShimWarning)

Traceback (most recent call last):
  File "/Users/aaron/Documents/virtualenvs/django18_py2/lib/python2.7/site-packages/django_extensions/management/commands/shell_plus.py", line 120, in get_notebook
    from IPython.frontend.html.notebook import notebookapp
ImportError: cannot import name notebookapp

Could not load IPython Notebook interactive Python environment.
```

tallus:
I put in a fix for this here: https://github.com/django-extensions/django-extensions/pull/723

You can get the kernel name using find_kernel_specs()

trbs:
Please try with 92008a3

joshowen:
That fixes the issue for me.

trbs:
great :)

[...]

PR #721, opened 2015-08-12
IPython 4.0.0 kernel spec name change

multigl:
python -> python2


multigl:
fixes #720

limnick:
+1

trbs:
I made another patch to solve this at 92008a3 which also gives the possibility to overwrite the list of kernel spec names to try via settings.py.

PR #722, opened 2015-08-13
Added ability to filter autocomplete query

zefciu:
It might be handy for the developer who creates an admin form to add some arbitrary additional filtering for the autocomplete field. An obvious use case would be to limit the list of linkable objects according to some per-row permission rule. This PR contains a simple method that allows to add arbitrary filter depending on the model and request.


trbs:
Thanks !

PR #723, opened 2015-08-14
 Fix for #720 iPython 4.0 breaks shell_plus notebook

tallus:
- Use ksm.find_kernel_specs() to get kernel name


trbs:
I've just commited 92008a3 which hopefully fixes this problem.

PR #730, opened 2015-08-28
Enable sqldsn with postgis engine

dpoirier:
This works fine once 'postgis' is added to the
check for compatible engines.


trbs:
Thanks !

PR #736, opened 2015-09-18
add globbing to extra files

bionikspoon:
# Example usage:

**Before:**

``` sh
$ django-admin runserver_plus --extra-file=app/templates/base.html --extra-file=app/templates/user_form.html
```

**After:**

``` sh
$ django-admin runserver_plus --extra-file=app/templates/*.html
```

or

``` sh
$ django-admin runserver_plus --extra-file=app/templates/*.html --extra-file=app/templates/layouts/*.html
```

or even

``` sh
$ django-admin runserver_plus --extra-file=app/templates/{*/,}*.html
```

ps.  Why am I watching HTML? crispy-forms.  There is a use case for this feature.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -240,7 +240,7 @@ class Command(BaseCommand):
             self.addr if not self._raw_ipv6 else '[%s]' % self.addr, self.port)
         # glob `--extra-file` input
         extra_files = list(set(
-            itertools. chain.from_iterable(glob.glob(nested_file)
+            itertools.chain.from_iterable(glob.glob(nested_file)
             for nested_file in options.get('extra_files', None) or [])
         ))
         reloader_interval = options.get('reloader_interval', 1)
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -1,4 +1,6 @@
 import logging
+import glob
+import itertools
 import os
 import re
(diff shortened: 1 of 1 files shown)


bionikspoon:
As a follow up suggestion:  create  a settings option for these extra watch files.

blueyed:
Nice, but consider quoting the glob pattern in the shell examples and the help (to be done): otherwise shells might complain, e.g. `no matches found` with Zsh (depending on options), because the shell will try to expand the glob (which included the option as a prefix).

And also with `--extra-file foo/*/*.py` it should not get expanded by the shell.

While at it, it would be nice to support `templates/**/*.html` (for all html files in all subdirs), but that's not possible with the standard lib IIRC, is it?!

bionikspoon:
As a follow up report, watching too many `.html` files reliably causes my computer to crash.  Not sure what's happening there, could be just my rig.  I'm just bringing this up as something to look out for.  It could have been the fact I  "ghetto forked" this library by copying and pasting the minimum amount into a command, so I could add the globbing.  This "ghetto fork" could have been part of the problem.

blueyed:
> watching too many .html files reliably causes my computer to crash
According to the werkzeug changelog, it does not use inotify anymore, but probably is polling instead.

You could use `strace -f` to see what's going on, and that might be the issue.

trbs:
@bionikspoon could you add a few lines to the documentation on using globs ?

camilonova:
@bionikspoon can you?

PR #747, opened 2015-10-13
Add value format for print_settings

bersace:
This allow to bridge django and bash scripts. Just call print_settings
with the wanted settings and get back the value.


trbs:
Thanks !

PR #751, opened 2015-10-20
Support for conditional expressions in shell_plus with Django >= 1.8

CesarBustios:
Conditional expressions are a nice new feature in Django >= 1.8. The commit just extends the import from ```django.db.models```


trbs:
Great addition, thanks !

CesarBustios:
You're welcome pal!

PR #754, opened 2015-10-26
Fixed reset_db with postgis in Django 1.9

artscoop:
With PostGIS 2+ and Django 1.9+, some steps are not required anymore or cause an ImportError. These commits fix the issue.


trbs:
I already merged #758, please open a new PR if there things left in this PR that should be merged.

PR #756, opened 2015-10-29
Update jquery 1.7.2 to 1.8.3.

nikolas:


trbs:
note: lets think about removing this all together. Django 1.5 support ended September 2, 2014.

camilonova:
@nikolas thank you, I'm going to remove that.

PR #764, opened 2015-11-17
Fix tests for Django master

ferrix:
There is a bug in pip that results in an unicode error. This is a workaround.

https://github.com/django/django/commit/bd059e3f8c6311dcaf8afe5e29ef373f7f84cf26#commitcomment-9896566


trbs:
Thanks!

PR #766, opened 2015-11-20
Added an option to disable the pin in werkzeug.

deese:
Disable the pin with the following parameter:

./manage.py runserver_plus --nopin

I know that can be insecure, but for local developing it helps a lot.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -1,21 +1,27 @@
+# -*- coding: utf-8 -*-
 import logging
 import os
 import re
 import socket
 import sys
 import time
-from optparse import make_option
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
 from django.db import connections, DEFAULT_DB_ALIAS
+from django.db.backends import utils
+from django.db.migrations.executor import MigrationExecutor
 from django.core.exceptions import ImproperlyConfigured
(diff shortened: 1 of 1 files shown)


trbs:
Could you make the PR such that it will work with both 0.10.x and 0.11.x ?
Also if you could add just a couple of lines in the documentation, about what this is ?
(Like a small paragraph explaining what this is and a link to the werkzeug docs about the pin security feature)

Thanks !

deese:
After reviewing the documentation, the pin can be disabled with an environment variable (http://werkzeug.pocoo.org/docs/0.11/debug/#debugger-pin) 

Do you think that this patch is needed? Maybe not.

I have an update of the patch, so if you think is something useful I can make the PR.

Thanks!

trbs:
I have no problem with the option in the command as long as it's backwards compatible with people running 0.10.x versions.

trbs:
bump, happy to merge that updated patch (assuming that it will be backwards compatible) if you want.

FeroxTL:
+1 for this option

deese:
I just changed the patch. As it's using the environment variable, it will work with any version.  Sorry for the delay, I just lost track of this :)

trbs:
Think the PR needs to be rebased to current master version...

deese:
I did it, but the PR is showing both commits. I will close this PR and create a new one.

Issue #708, opened 2015-07-27
Django 1.9 Exception: can't adapt Type JSONDict


jaddison:
**Edit: confirmed that this is still an issue in the `master` branch as of today.**

I realize Django 1.9 isn't out yet - but in working with it, I've discovered that recent changes  in django cause the following traceback relating django-extensions. I use `JSONField()` on one of my models and try to `get_or_create` that model, passing in no data for the json field.

Perhaps related to `psycopg2`'s http://initd.org/psycopg/docs/faq.html#problems-with-type-conversions type adapting requirements? 

```
Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/proj/src/django/django/db/models/query.py", line 463, in get_or_create
    return self.get(**lookup), False
  File "/home/vagrant/.virtualenvs/proj/src/django/django/db/models/query.py", line 387, in get
    self.model._meta.object_name
users.models.social.DoesNotExist: SocialConnection matching query does not exist.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/vagrant/.virtualenvs/proj/src/django/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
psycopg2.ProgrammingError: can't adapt type 'JSONDict'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "./manage.py", line 25, in <module>
    execute_from_command_line(sys.argv)
  File "/home/vagrant/.virtualenvs/proj/src/django/django/core/management/__init__.py", line 331, in execute_from_command_line
    utility.execute()
  File "/home/vagrant/.virtualenvs/proj/src/django/django/core/management/__init__.py", line 323, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/vagrant/.virtualenvs/proj/src/django/django/core/management/base.py", line 350, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/vagrant/.virtualenvs/proj/src/django/django/core/management/base.py", line 401, in execute
    output = self.handle(*args, **options)
  File "/vagrant/utils/management/commands/migrate_proj.py", line 140, in handle
    self.migrate_users(conn)
  File "/vagrant/utils/management/commands/migrate_proj.py", line 123, in migrate_users
    token=facebook_token
  File "/home/vagrant/.virtualenvs/proj/src/django/django/db/models/manager.py", line 122, in manager_method
[...]

PR #772, opened 2015-12-04
Django 1.9 compatibility

nikolas:
Also enabling automated testing on django 1.9 and python 3.5


maryokhin:
Hi, @trbs! Really interested for @nikolas's fixes to go in. 
Do you guys maybe need some help to speed up the merge/release of this?

django_extensions/management/commands/show_templatetags.py
```diff
@@ -93,14 +97,22 @@ def handle(self, *args, **options):
                 app_labeled = False
                 for taglib in tag_files:
                     try:
-                        lib = get_library(taglib)
+                        try:
+                            lib = get_installed_libraries()[taglib]
```
nikolas: @maryokhin there's just one failing test remaining on this branch and it has to do with the show_templatetags command here. I've made django-1.9 compatible code here, with the deprecation of `get_library`, and now the issue is that the `django_extensions` template tag library doesn't seem to be loaded during testing in django1.9 here. Maybe someone can fix the problem? Otherwise I will look again later.

trbs:
It's a big patch and I would love to get it in.

More eyes on it will definitely help :)

trbs:
looks like you are only missing one import:

```
diff --git a/django_extensions/management/commands/show_templatetags.py b/django_extensions/management/commands/show_templatetags.py
index d558a24..6f1d2ed 100644
--- a/django_extensions/management/commands/show_templatetags.py
+++ b/django_extensions/management/commands/show_templatetags.py
@@ -1,6 +1,6 @@
 import inspect
 import os
-
+import importlib
 import six
 from django.core.management import color
@@ -99,7 +99,7 @@ class Command(BaseCommand):
                     try:
                         try:
                             lib = get_installed_libraries()[taglib]
-                            lib = import_module(lib).register
+                            lib = importlib.import_module(lib).register
                         except NameError:
                             lib = get_library(taglib)
                     except:
```

However I would prefer if we don't use to many try-except blocks instead we can explicitly check the Django version number. Possibly wrap the code that differs in some functions so it's transparant for the rest of the code.

That will avoid errors where exceptions are swallowed, hidden or otherwise creating new bugs..

nikolas:
Thanks! I added a new commit. Yeah, the nested try/except block in that particular place is a little crazy. I'll see if I can clean it up.

My instincts are to try to avoid checking Django version numbers, but maybe it's fine, especially if it's necessary in order to make the code clearer.

maryokhin:
Just a suggestion, this is how a big Django-dependent framework handles this, in `compats.py`:
https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/compat.py

nikolas:
Alright, I've moved the code you mentioned into compat.py, and in my mind the big problem was the bare `except:` statement, swallowing valid exceptions like the missing import.

I've qualified this statement with `(InvalidTemplateLibrary, KeyError)`, catching only the exceptions that signal a missing template library, which we then use to `continue`. I guess the new `compat.load_tag_library` function could maybe be made clearer by looking at the django version, so if I come up with something I like more I will add a new commit.

nikolas:
Here's an alternative `load_tag_library` function that checks the django version instead of using a nested try/except. Let me know if you like this one more or have other suggestions:

```py
import django


def load_tag_library(libname):
    """Load a templatetag library on multiple Django versions.

    Returns None if the library isn't loaded.
    """
    if django.VERSION < (1, 9):
        from django.template.base import get_library, InvalidTemplateLibrary
        try:
            lib = get_library(libname)
            return lib
        except InvalidTemplateLibrary:
            return None
    else:
        import importlib
        from django.template.backends.django import get_installed_libraries
        from django.template.library import InvalidTemplateLibrary
        try:
            lib = get_installed_libraries()[libname]
            lib = importlib.import_module(lib).register
            return lib
        except (InvalidTemplateLibrary, KeyError):
            return None
```

trbs:
Yes, I would prefer the latter one.

Can we update the other places where we need this for Django 1.9 compatibility as well ?

nikolas:
Sure, I'll do that tonight or tomorrow.

nikolas:
Alright this has been cleaned up and rebased, with a lot of the compatibility code now in `compat.py`.

trbs:
Amazing ! Thanks for your hard work !

One question and One request:

Question: In django_extensions/db/fields/json.py does moving the ```super()``` call up change the behavior of the field ? and if so is this intended (bugfix) or not.

Request: Please add your name to the docs/AUTHORS list if you want :)

nikolas:
In fields/json.py, I've changed the time at which the data is processed when going into the database, from the `get_db_prep_save` hook to `get_db_prep_value`, following this example: https://docs.djangoproject.com/en/1.9/howto/custom-model-fields/#converting-query-values-to-database-values

This fixes a test failure that occurs in Django 1.9, seen here:
https://travis-ci.org/django-extensions/django-extensions/jobs/94802320#L209

It shouldn't change the behavior of the field, but I have to admit I'm not that familiar with custom fields... what I've done seemed like the right thing to do, and if fixes the test, so that's all I can say.

jaddison:
@trbs the `json.py` change by @nikolas should actually fix the bug I submitted here: https://github.com/django-extensions/django-extensions/issues/708

jaddison:
Confirmed, installing @nikolas' branch fixes #708.

trbs:
@jaddison cool thanks for checking that !

trbs:
Thanks for all the work on this !

Please all test this as much as possible.
I will make a release in a couple of days if there are no issues.

brianmay:
Not sure if the fixes here made it into 1.6.1 or not, however I am still getting test failures with version 1.6.1

nikolas:
Yes, this patch was merged here: https://github.com/django-extensions/django-extensions/commit/8ef6f5e659b36a92b8dbb6ab3a1f3b735f84272a

@brianmay Are you getting different results on your computer than what travis is showing? https://travis-ci.org/django-extensions/django-extensions

brianmay:
I posted a log to #796 before closing it. That log was produced with the latest release, so I wondered if something had happened after the latest release to fix the problems.

jaddison:
1.6.0 apparently fixed issues with jsonfield, but didn't really. 1.6.1 was issued to undo those changes.

https://github.com/django-extensions/django-extensions/issues/781

PR #774, opened 2015-12-07
Better emulate read_default_file DB config parsing

### LINKED ISSUE #773: read_default_file not handled correctly
read_default_file is used by reset_db and drop_test_database

They process read_default_file themselves differently from the way django does, specifically:
- Does not expand user home directories
- Expects that a [client] section will be present if read_default_file is specified and throws an exception if not
- Expects all of user, password, database will be present if read_default_file is specified and throws an exception if one or more is not present
- Ignores any socket/host/port options in the config file

Incidentally, sqlcreate() and sqldsn() have their own handling that appears to not use read_default_file at all (although they don't make DB changes so are less likely to do unexpected things)

levic:
Fixes #773


levic:
Arguably all of the database config setting parsing code (not just that for mysql, and not just these two commands) should be unified into a single place, but that's a much more extensive change with greater risk of breaking other things so I went with the simpler option.

trbs:
Could you please move the ```management/commands/_private.py``` file to ```management/mysql.py``` ?
That way we don't have none-command files in the ```commands/``` directory.

trbs:
Thanks !

PR #777, opened 2015-12-09
Fixed: ForeignKey admin called deprecated method

ahankinson:
When using the ForeignKeyAutocompleteAdmin field, the search-as-you-type functionality was calling a method, `models.get_model` that was removed in Django 1.9.

This commit fixes this by importing the django 'apps' model and replacing it with `apps.get_model`.

### DIFF
--- a/django_extensions/admin/__init__.py
+++ b/django_extensions/admin/__init__.py
@@ -9,6 +9,7 @@ import django
 from django.http import HttpResponse, HttpResponseNotFound
 from django.conf import settings
 from django.db import models
+from django.apps import apps
 from django.db.models.query import QuerySet
 from django.utils.encoding import smart_str
 from django.utils.translation import ugettext as _
@@ -102,7 +103,7 @@ class ForeignKeyAutocompleteAdmin(ModelAdmin):
                     return "%s__search" % field_name[1:]
                 else:
                     return "%s__icontains" % field_name
-            model = models.get_model(app_label, model_name)
+            model = apps.get_model(app_label, model_name)
             queryset = model._default_manager.all()
             data = ''
             if query:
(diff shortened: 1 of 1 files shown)


trbs:
Are we sure this work for our compatibility policy ? 
https://django-extensions.readthedocs.org/en/latest/#compatibility-with-versions-of-python-and-django

ahankinson:
I believe `apps.get_model` was introduced in Django 1.7; it's the earliest reference I can find. I can probably add a check if that's what you would like. 

However, your compatibility page says two different things. At the top it says you follow the guidelines here: https://www.djangoproject.com/download/#supported-versions and that you don't guarantee that django extensions would work with non-supported Django versions. Django 1.7 went out of extended support on December 1st 2015. However, you then go on to say that you support versions as far back as 1.4. (Personally, I suspect that few people are expecting the newest version of Django Extensions to work on a much older version of Django, but I may be wrong about that...)

Could you clarify, then, whether you want it to work with 1.4+, or if you want it to work with supported Django versions? (Currently 1.8 & 1.9)

trbs:
I found that too when I was looking it up for my comment :)
And you are right about the inconsistency in the docs.

The first rule of thumb is that ```django-extensions``` will only support releases that are currently supported by Django itself. (Of course as we might lag a bit behind Django's official releases, I try to minimize that but with time, people and eventually money being a limited commodity we cannot just always guarantee that we will be on par with Django releases as they come out, however I try to get there as soon as possible)

The second rule is that I like to try to be as backwards compatibility as we can as long as it's not violating the first rule or becoming a massive hassle for maintenance.

If you could add a try-except clause to fallback for older version that would be much appreciated :)

ahankinson:
Fixed. Let me know if this is acceptable.

trbs:
Thanks !

Issue #770, opened 2015-12-03
Use of option_list in Commands triggers deprecation warning in Django 1.9


jonblum:
Running `runserver_plus.py` and most other django-extensions commands produces the following deprecation warning:
```
11:26:59 web.1  | WARNING  [py.warnings:265] {snip}/lib/python2.7/site-packages/django/core/management/base.py:265:
RemovedInDjango110Warning: OptionParser usage for Django management commands is deprecated, use ArgumentParser instead
```

[The docs](https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/#django.core.management.BaseCommand.option_list) advise using `add_arguments` in its place:
> You should now override the add_arguments() method to add custom arguments accepted by your
> command. See the example above.

This will be breaking in Django 1.10.

thecardcheat:
Duplicate of #788?

PR #785, opened 2015-12-17
Fixes RemovedInDjango110Warning in runserver_plus

heyman:
This fixes the issue described in #770, but only for the runserver_plus command.


heyman:
I *think* that the failing tests are unrelated to my changes.

PR #791, opened 2015-12-28
Fixed 'Error binding parameter 2' in Django 1.9

szok:


trbs:
The last patch for this caused side effects... can people test this (and/or create more tests) so we are sure this works correctly for everybody ?

szok:
I tested only for Django 1.9.
According to unit testing should work other version

trbs:
The unit tests are not covering all cases. There is a PR for some additional tests here: https://github.com/django-extensions/django-extensions/pull/793

trbs:
I do not want to merge anything that is not thoroughly tested....

szok:
Ok, tests added.

brianmay:
What is the status of this pull request? The tests appear to pass after installing making this change.

trbs:
Think the status is still the same... Either somebody needs to be able to explain why this change is correct and why it won't break anything and/or it needs to be thoroughly tested.

brianmay:
I assume you mean the test cases already added by @szok on 5th Jan are insufficient?

camilonova:
@szok Can you please check the conflicts?

szok:
@camilonova Yes, it will be on Monday

szok:
@camilonova Done :)

camilonova:
@szok Thanks man

Issue #752, opened 2015-10-22
RandomCharField error when saving Model


PawelDecowski:
``` python
class Test(models.Model):
    code = RandomCharField(length=6, include_alpha=False)

test = Test()
test.save()
```

> DataError at /account/password/
> value too long for type character varying(6)

After further inspection:

``` sql
INSERT INTO "test_test" ("code") VALUES (\'<generator object random_char_generator at 0x7ffea2080f00>\') RETURNING "test_test"."id"
```

I also tried this:

```
>>> c = RandomCharField(length=6, include_alpha=False)
>>> c.random_char_generator('0123456789')
<generator object random_char_generator at 0x7f32ac4aec80>
```

What’s happening?

PR #792, opened 2015-12-28
Return the first random string if RandomCharField is not unique (#752).

fladi:
`random_chars` is a generator so we need to take the first value from it if we
do not care about uniqueness.


trbs:
Thanks !

Do you think you could add a test for this as well ?

PR #793, opened 2015-12-28
Add separate tests for non-unique RandomCharField.

fladi:
This adds a test for #792 which covers a RandomCharField with `unique=False`.


trbs:
Besides the new test, other tests now also become ```RandomCharTestModelUnique``` is this intentional ?

PR #797, opened 2016-01-08
Django 1.9 compatibility

hugotacito:
Changes the clean_pyc command to be compatible with django 1.9


trbs:
afaik this breaks both backwards compatibility for arguments as well as the signals for commands.

Issue #789, opened 2015-12-21
reset_db: Raise CommandError when failing to delete sqlite database file


ztarem:
I use the reset_db command inside a larger script that recreates the database on demand. When the database is open somewhere else (an IDE, a database browser, etc.) the command fails silently. Raising an exception would be very useful.

trbs:
Closing this for being inactive for too long. If there is a better way we are still happy to accept a PR for this.

PR #800, opened 2016-01-12
Raise CommandError when failing to remove the sqlite database file

ztarem:
Proposed fix for issue #789: "reset_db: Raise CommandError when failing to delete sqlite database file".


trbs:
Seems like a good idea...

Could you add an option to '--fail-silently', so people can have the old behavior if they really want/need too ?

Thanks

PR #803, opened 2016-01-14
Fix compatibility with Django 1.10+

Stranger6667:
`SubfieldBase` is deprecated since Django 1.10


Stranger6667:
It definitely mistake

Issue #788, opened 2015-12-20
Django >= 1.8 ; Deprecated OptParser usage ArgParse; Warnings on running runserver_plus


sudheesh001:
Running `runserver_plus` with `Werkzeug` debugger throws the following warning with the latest Django

```
RemovedInDjango110Warning: OptionParser usage for Django management commands is deprecated, use ArgumentParser instead
  RemovedInDjango110Warning)
```

Looks like this is happening because of the usage of `*args` and `**options`,  See the changed in 1.8 which says https://docs.djangoproject.com/en/1.9/howto/custom-management-commands/#module-django.core.management
```
Changed in Django 1.8:
Before Django 1.8, management commands were based on the optparse module, and positional 
arguments were passed in *args while optional arguments were passed in **options. Now that 
management commands use argparse for argument parsing, all arguments are passed in **options 
by default, unless you name your positional arguments to args (compatibility mode). You are 
encouraged to exclusively use **options for new commands.
```

nehemiahjacob:
+1

thecardcheat:
Please see my #804.

thecardcheat:
#804 should close this.

timgraham:
`runscript` is still using `optparse`, and I'm not sure how to convert the `vararg_callback` part of it. Here's a start:
```diff
diff --git a/django_extensions/management/commands/runscript.py b/django_extensions/management/commands/runscript.py
index f7c01f0..a14b1a4 100644
--- a/django_extensions/management/commands/runscript.py
+++ b/django_extensions/management/commands/runscript.py
@@ -1,8 +1,8 @@
 # coding=utf-8
+import argparse
 import importlib
 import sys
 import traceback
-from optparse import make_option
 
 from django.apps import apps
 
@@ -11,39 +11,36 @@ from django_extensions.management.email_notifications import \
 from django_extensions.management.utils import signalcommand
 
 
-def vararg_callback(option, opt_str, opt_value, parser):
-    parser.rargs.insert(0, opt_value)
-    value = []
-    for arg in parser.rargs:
-        # stop on --foo like options
-        if arg[:2] == "--" and len(arg) > 2:
-            break
-        # stop on -a like options
-        if arg[:1] == "-":
-            break
-        value.append(arg)
+class VarArgCallback(argparse.Action):
 
-    del parser.rargs[:len(value)]
-    setattr(parser.values, option.dest, value)
+    def __call__(self, parser, namespace, values, option_string=None):
[...]

Issue #796, opened 2016-01-03
Tests fail for Django 1.9


brianmay:
```
I: pybuild base:184: python2.7 /usr/bin/py.test
/usr/lib/python2.7/dist-packages/django/db/models/fields/subclassing.py:22: RemovedInDjango110Warning: SubfieldBase has been deprecated. Use Field.from_db_value instead.
  RemovedInDjango110Warning)

============================= test session starts ==============================
platform linux2 -- Python 2.7.11, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
rootdir: /«PKGBUILDDIR», inifile: setup.cfg
plugins: cov-2.2.0, django-2.9.1
collected 67 items

tests/test_autoslug_fields.py .........
tests/test_clean_pyc.py ...
tests/test_compile_pyc.py ..
tests/test_dumpscript.py ....
tests/test_encrypted_fields.py ssssssss
tests/test_json_field.py FFF
tests/test_management_command.py .....
tests/test_models.py ......
tests/test_randomchar_field.py ..........
tests/test_runscript.py ..
tests/test_shortuuid_field.py ....
tests/test_templatetags.py ..
tests/test_utils.py ....
tests/test_uuid_field.py .....

=================================== FAILURES ===================================
_____________________ JsonFieldTest.test_char_field_create _____________________

self = <tests.test_json_field.JsonFieldTest testMethod=test_char_field_create>

    def test_char_field_create(self):
>       j = JSONFieldTestModel.objects.create(a=6, j_field=dict(foo='bar'))

tests/test_json_field.py:8: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/lib/python2.7/dist-packages/django/db/models/manager.py:122: in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
/usr/lib/python2.7/dist-packages/django/db/models/query.py:401: in create
    obj.save(force_insert=True, using=self.db)
/usr/lib/python2.7/dist-packages/django/db/models/base.py:700: in save
    force_update=force_update, update_fields=update_fields)
/usr/lib/python2.7/dist-packages/django/db/models/base.py:728: in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
[...]

PR #804, opened 2016-01-15
Transition command files to argparse compatibility.

thecardcheat:
Issue #788.

I wanted feedback on this patch regarding a few things before putting more time in:

1) Are you cool with this method? This seems a lot more sensible to me than maintaining two argument lists in every management command file. It seems that BaseCommand simply uses the presence of an option_list to determine what course to take.

2) In my _extremely_ brief testing, everything seemed kosher in 1.9.1 and 1.7.11 using actual types for the type kwarg in optparse (e.g. the indent arg in print_settings; this is what argparse wants), despite the optparse doc only mentioning strings. Anyone know the deal here? Please point out any major pitfalls you see that I might have missed.


thecardcheat:
Also for clarity because I allowed the pull request title to be misleading, my intent was to do this project-wide - I just wanted initial feedback.

trbs:
Yes, I think this is a clean way of doing it. It's very nice that we can use the ```def add_arguments``` everywhere and still be backwards compatible. Also removing this code in the future will be straight forward.

I think we can deal with any pitfall that might arise. Also Django 1.7 is now officially unsupported since 1 December 2015. So we don't have to be super strict about supporting it, although I always like it when we can be friendly to previous versions :)

Please update the patch and I will merge it in.
Thanks for your work on this !

thecardcheat:
I think that about covers it. I am guessing that those 1.9 failures relate to #796? If you can think of anything else that will help this particular PR, let me know. Also, added myself to AUTHORS because I stumbled across it and could not find directions for/against it.

trbs:
Thanks ! Yeah the 1.9 failures are related to our ```JSONField``` most important for that is to have a patch that works and not breaks anything :) Regretfully I do not have the time currently to really dive into the patches and work out what exactly is happening there, what changes and what a correct patch should look like... then the next best thing is having it heavily tested by users...

EduardoCastanho:
hi ! 
I am using the github master version to run this command :
`python manage.py graph_models -a -g -o my_project_visualized.png `

But i am getting the folling error:
```
`Traceback (most recent call last):
  File "manage.py", line 28, in <module>
    execute_from_command_line(sys.argv)
  File "/home/vagrant/.venvs/geosea/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/vagrant/.venvs/geosea/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/vagrant/.venvs/geosea/local/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/vagrant/.venvs/geosea/local/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/home/vagrant/django-extensions/django_extensions/management/utils.py", line 58, in inner
    ret = func(self, *args, **kwargs)
  File "/home/vagrant/django-extensions/django_extensions/management/commands/graph_models.py", line 91, in handle
    self.options_from_settings(options)
  File "/home/vagrant/django-extensions/django_extensions/management/commands/graph_models.py", line 120, in options_from_settings
    for option in self.graph_models_options:
AttributeError: 'Command' object has no attribute 'graph_models_options'
```

I check the code  and it seems to me that the method 'options_from_settings()' was forgotten in the solution to this issue.

Can you check this ?

Thanks

thecardcheat:
Thanks for report. I have opened #835.

PR #815, opened 2016-01-25
Update admin_generator for Django 1.9. Fix #784

### LINKED ISSUE #784: admin_generator broken in 1.9
Running the admin_generator command gives me the following error:

`LookupError: No installed app with label '<module 'myapp.models' from '/path/to/myapp/models.pyc.`

It looks like the command is trying to use the repr of the models module, rather than the app name.

I'm using 1.6.1 BTW

readevalprint:
Fixes the error where it expects an app name but gets the module instance.

```
LookupError: No installed app with label '<module 'my_app.models' from '/path/to/my_app/models.py'>'.
```


trbs:
Could you please fix the test errors ?
(Sorry for the inconvenience since one other test is also failing)

Also if you could please add a test case for this ?

readevalprint:
Ah yes I was sick and then totally forgot about this.

trbs:
Think we already have another fix for this which still allows running the command without any arguments.

If this didn't solve the issue please open a new PR.

Issue #787, opened 2015-12-18
Werkzeug debugger pin doesn't appear when using runserver_plus command


YPCrumble:
When I use `runserver_plus` and try to use the interactive console, I get this warning:

> Console Locked
> The console is locked and needs to be unlocked by entering the PIN. You can find the PIN printed out on the standard output of your shell that runs the server.

But the console only shows:

> Performing system checks...

> System check identified no issues (0 silenced).

> Django version 1.9, using settings 'settings.local'
> Development server is running at http://127.0.0.1:8000/
> Using the Werkzeug debugger (http://werkzeug.pocoo.org/)
> Quit the server with CONTROL-C.

sealgair:
I had this same problem, but it turned out I'd disabled the werkzeug logger.  I just added the following to LOGGING['loggers'] in settings.py:

        'werkzeug': {
            'handlers': ['console'],
            'level': 'DEBUG',
            'propagate': True,
        }

trbs:
Anybody have a little of time to add this example to the documentation ? :)

PR #821, opened 2016-02-04
Small update to `runserver_plus` docs for logging settings

linssen:
Ran into #787, and this sorted it. This felt like the right place for it, and the right format, hope it's ok.


camilonova:
Thank you

PR #822, opened 2016-02-05
Updated the runscript docs to match the Django tutorial for Django 1.8+

alasdairnicol:
The current example was causing a bit of confusion on [this stack overflow question](http://stackoverflow.com/questions/35226883/python-django-typeerror-relative-imports-require-the-package-argument/35227881?noredirect=1#comment58173246_35227881), so I've updated the example to more closely match the Django tutorial for Django 1.8+.


camilonova:
Thank you

PR #824, opened 2016-02-07
Improve profile doc

rthouvenin:
While trying the *runprofileserver* command, I got an error while opening the profiling results.
After some investigation I found out that this is because I was using the default profiler (hotshot) as opposed to cProfile.

I am submitting this PR to improve the documentation in the hope that this will help someone with the same issue. I also slightly modified the code of the command to replace "hotsPot" instances with "hotsHot", this the real name is "hotshot" and it may make it easier for someone searching the word.


camilonova:
Thank you

PR #828, opened 2016-02-09
Make ForeignKeySearchInput usable outside changeforms

alexmorozov:
Previously the `related_url` variable was calculated assuming the page with a widget is nested 3 levels down (like a model's changeform, `/app/model/<pk>`). That was making it harder to use the widget in a custom admin views, which could reside at the arbitrary nesting level. Here we switch to using a more flexible `reverse()` call.

### DIFF
--- a/django_extensions/admin/widgets.py
+++ b/django_extensions/admin/widgets.py
@@ -1,9 +1,9 @@
 # coding=utf-8
 import six
 from django import forms
-from django.core.urlresolvers import reverse
 from django.contrib.admin.sites import site
 from django.contrib.admin.widgets import ForeignKeyRawIdWidget
+from django.core.urlresolvers import reverse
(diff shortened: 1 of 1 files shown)


alexmorozov:
@trbs any chance of getting this merged? Or does it break any backwards compatibility (should not, actually)?

trbs:
@alexmorozov thanks for the bump and PR. I'm sorry for the long delay... been very busy.

trbs:
Github says that: ```This branch has conflicts that must be resolved```.

Could you try and update/pull your branch ? hopefully I can merge it then.

alexmorozov:
Sure thing. It says coverage decreased, but seems fine overall, tests pass. Check this out, please.

trbs:
Thanks !

PR #831, opened 2016-02-11
shell_plus dynamic ipython args from command instance

bshelton229:
I need to set some ipython args that are unique to each command run, specifically `--logfile`. This should let subclasses override ipython args per instance. I couldn't think of a better way to get this working.


trbs:
Thanks !

Issue #799, opened 2016-01-12
graph-models fails in python3 (TypeError: can't concat bytes to str)


haakenlid:
I just moved a project from python 2.7 to python 3.4 and graph-models fails with this error message.
```
TypeError: can't concat bytes to str   
File "/srv/env/lib/python3.4/site-packages/django_extensions/management/modelviz.py", 
line 216, in add_relation   
label += ' (%s)' % related_query_name
```

haakenlid:
Are there any existing tests for the graph_models management command?

PR #834, opened 2016-02-17
Fix bug with graph_models in python 3

haakenlid:
Fixes issue #799
I don't know if this needs a test or if it's fine the way it is.


trbs:
Preferably every needs tests :)
If you could add a test that would be wonderful.
Thanks for the patch !

Issue #820, opened 2016-02-02
./manage.py admin_generator: error: unrecognized arguments: ... and some exceptions...


prokher:
I would like to generate admin interface for my apps. First I run:
```bash
> ./manage.py admin_generator
This command requires an existing app name as argument
Available apps:
    apps.portal.api
    debug_toolbar
    django.contrib.admin
    django.contrib.auth
    django.contrib.contenttypes
    django.contrib.sessions
    django_extensions
    rest_framework
```

Here I see the list of apps and try to generate_admin for one of them:
```bash
> ./manage.py admin_generator apps.portal.api
usage: manage.py admin_generator [-h] [--version] [-v {0,1,2,3}]
                                 [--settings SETTINGS]
                                 [--pythonpath PYTHONPATH] [--traceback]
                                 [--no-color] [-s SEARCH_FIELD]
                                 [-d DATE_HIERARCHY] [-p PREPOPULATED_FIELDS]
                                 [-l LIST_FILTER_THRESHOLD]
                                 [-r RAW_ID_THRESHOLD]
manage.py admin_generator: error: unrecognized arguments: apps.portal.api
```

I thought I did something wrong and run
```bash
> ./manage.py admin_generator --help
Traceback (most recent call last):
  File "./manage.py", line 8, in <module>
    execute_from_command_line(sys.argv)
  File "/home/prokher/Dev/PyVirtEnvs/pSeven/lib/python3.4/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/prokher/Dev/PyVirtEnvs/pSeven/lib/python3.4/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/prokher/Dev/PyVirtEnvs/pSeven/lib/python3.4/site-packages/django/core/management/base.py", line 339, in run_from_argv
    options = parser.parse_args(argv[2:])
  File "/home/prokher/Dev/PyVirtEnvs/pSeven/lib/python3.4/site-packages/django/core/management/base.py", line 59, in parse_args
    return super(CommandParser, self).parse_args(args, namespace)
  File "/usr/lib/python3.4/argparse.py", line 1737, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/usr/lib/python3.4/argparse.py", line 1769, in parse_known_args
    namespace, args = self._parse_known_args(args, namespace)
[...]

PR #838, opened 2016-02-18
Fix admin_generator incompatible "default" identifier raising TypeError.

thecardcheat:
In admin_generator.py we change to the argparse "default" string replace
style and update ProxyParser to handle the necessary replacement for
optparse. The 3.5.1 docs for argparse state: "The help strings can
include various format specifiers to avoid repetition of things like the
program name or the argument default. The available specifiers include
the program name, %(prog)s and most keyword arguments to add_argument(),
e.g. %(default)s, %(type)s, etc." The use of "most" is ambiguous and in
preliminary testing, this functionality is at least missing for the
kwarg "action". It is not the intent of CompatibilityBaseCommand or
ProxyParser to add functionality to optparse, so we make no effort to
handle whatever other kwargs may be supported as formatted
identifiers in argparse as the optparse docs only mention %default
in the help string. %prog exists in optparse only in the usage and
version strings.

### DIFF
--- a/django_extensions/compat.py
+++ b/django_extensions/compat.py
@@ -206,7 +206,16 @@ class ProxyParser(object):
         self.command = command
 
     def add_argument(self, *args, **kwargs):
-        self.command.option_list +=  (make_option(*args, **kwargs), )
+        """Transform our argument into an option to append to self.option_list.
+
+        In argparse, "available specifiers [in help strings] include the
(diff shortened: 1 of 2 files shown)


thecardcheat:
Resolve #820

marctc:
Nice! Glad to see you find the solution quick!

trbs:
Thanks !

Issue #835, opened 2016-02-17
Using the GRAPH_MODELS setting is broken in graph_models command


thecardcheat:
This is due to the change from 160aaa80829a9e2cd45ff00d7054e9bb09393b98 that introduced CompatibilityBaseCommand. In the graph_models.py Command, options_from_settings() expects self.graph_models_options, which was the list of options originally appended to BaseCommand.option_list. For the condition where self.option_list exists, it should be a simple replace to the current options_from_settings(). For argparse, the change required needs to be assessed.

thecardcheat:
I will take this.

PR #840, opened 2016-02-19
Fix settings.GRAPH_MODELS breaking graph_models.

thecardcheat:
I considered a couple different ways to do this and this seemed the most appropriate considering that CompatibilityBaseCommand should be considered temporary until support for Django 1.8 is removed (so not putting extra non-compat-specific code in there), removing potentially fragile library-specific code rather than adding more is better for future maintenance, and abstracting out settings defaults from this script is premature and may be appropriate as a feature addition, but not to resolve this issue.

### DIFF
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -1,6 +1,5 @@
 # coding=utf-8
 import sys
-from optparse import NO_DEFAULT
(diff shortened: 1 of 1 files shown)


thecardcheat:
Resolve #835

trbs:
Thanks !

PR #843, opened 2016-02-25
docs: add update_permissions command

Fak3:
Mention update_permissions management command in list of all command extensions


trbs:
Thanks

PR #845, opened 2016-02-28
Prevent 2nd reload of debugger on runserver_plus

### LINKED ISSUE #832: runserver_plus creates two pin codes for werkzeug debugger
Whenever I use runserver_plus to start a development server using the werkzeug debugger, it activates the debugger twice and prints out two pin codes. I expect there to be only one. Other than printing two pin codes in the terminal, the server runs perfectly as expected.

Steps to replicate:

```
# create a new empty virtualenv
pip install django ==1.9.2
pip install django-extensions==1.6.1
pip install werkzeug==0.11.3
django-admin.py startproject testproj
# Edit the testproj INSTALLED_APPS setting to include django_extensions
./manage.py runserver_plus
```

Output looks like this.
```
...
 * Debugger is active!
 * Debugger pin code: 140-764-085
 * Debugger is active!
 * Debugger pin code: 895-622-303
```

tony:
Only use the DebuggedApplication hander-wrapper on the outer call.

For manageability, move the inner_run into its own instance method.

Fixes #832
Related http://stackoverflow.com/a/9476701

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -340,6 +340,12 @@ class Command(BaseCommand):
             else:
                 extra_files.extend(filter(lambda filename: filename.endswith('.mo'), gen_filenames()))
 
+        # Werkzeug needs to be clued in its the main instance if running
+        # without reloader or else it won't show key.
+        # https://git.io/vVIgo
+        if not use_reloader:
+            os.environ['WERKZEUG_RUN_MAIN'] = 'true'
+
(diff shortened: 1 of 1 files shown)


tony:
Rebased after 1.6.6 release @ 9a16c06 , would like to get a QA on this approach.

tony:
Whats your take on this, @trbs?

This is a fix for a bug where the keycode shows twice on ``runserver_plus``.

screen shot 2016-05-10 at 7 06 09 am

trbs:
Looking good, thanks for your work on this !

PR #846, opened 2016-02-28
Fix call grind format to enable source code navigation in qcachegrind.

shuoli84:
ref: https://kcachegrind.github.io/html/CallgrindFormat.html

The filename should be in a separated line e.g "fl", "cfl".

### DIFF
--- a/django_extensions/management/commands/runprofileserver.py
+++ b/django_extensions/management/commands/runprofileserver.py
@@ -26,8 +26,6 @@ except ImportError as e:
     USE_STATICFILES = False
 
 
-
-
 class KCacheGrind(object):
     def __init__(self, profiler):
         self.data = profiler.getstats()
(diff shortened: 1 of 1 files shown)


shuoli84:
Here is the screenshot after the fix.
screen shot 2016-02-28 at 10 20 07 pm

shuoli84:
Open an issue related with this pull request. issue #847

trbs:
Is this a change in the format or a straight out bug ?

Aka could this break people (now) working setups ?

shuoli84:
@trbs I treat it as a bug, since for current version, I can't make qcachegrind work properly. But not sure whether some other people already rely on current format. Anyway, this only affect the callgrind profile, so should be ok.

trbs:
Thanks for the patch.

Issue #740, opened 2015-09-29
graph_models fails in Python 3.x


JoseTomasTocino:
Looks like trying to graph the built-in `django.contrib.auth` models in a fresh install with Python 3.x fails due to string-handling issues. **Testcase**:

```
mkdir testcase
pyvenv-3.4 myenv
source myenv/bin/activate
pip install django django_extensions pygraphviz
django-admin startproject testproject .
```

Then add `django_extensions` to `INSTALLED_APPS` and run:

```
./manage.py graph_models -g -o myapp_models.png -R -S -n -a
```

This appears:

``` text.python.traceback
Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django/core/management/__init__.py", line 330, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django/core/management/base.py", line 393, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django/core/management/base.py", line 444, in execute
    output = self.handle(*args, **options)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django_extensions/management/utils.py", line 57, in inner
    ret = func(self, *args, **kwargs)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django_extensions/management/commands/graph_models.py", line 77, in handle
    dotdata = generate_dot(args, cli_options=cli_options, **options)
  File "/home/jose/test333/myenv/lib/python3.4/site-packages/django_extensions/management/modelviz.py", line 141, in generate_dot
    model['label'] = appmodel._meta.verbose_name.decode("utf8")
AttributeError: '__proxy__' object has no attribute 'decode'
```

trbs:
Hope this is solved with the above PR's :)

PR #851, opened 2016-03-07
Python3 fixes to graph_models: byte/str coercion and version comparison

yourcelf:
Addresses two issues that prevent `graph_models` from running under python3:

 - `TypeError: unorderable types: tuple() < float()`: The version check for Django 1.7 uses a float instead of a tuple, this is an invalid comparison in python3.
 - `TypeError: Can't convert 'bytes' object to str implicitly` (issue #740).  Instead of relying on `+=` to do implicit str conversion, use a formatting string (`"{} ({})".format(...)`).  While the fix in this PR enables the command to run successfully, it's not ideal, as under python3 it will retain the `b''` prefix on the resulting string.  Nevertheless, it's better than a crash.

### DIFF
--- a/django_extensions/compat.py
+++ b/django_extensions/compat.py
@@ -200,7 +200,7 @@ def add_to_builtins_compat(name):
 
 
 def get_model(path):
-    if django.VERSION < (1.7):
+    if django.VERSION < (1, 7):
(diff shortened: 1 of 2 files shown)


trbs:
Thanks !

PR #867, opened 2016-04-11
Enable aliased imports in shell_plus pre- and post-imports

MattFisher:
Basically I just wanted 'from pprint import pprint as p'

### DIFF
--- a/django_extensions/management/shells.py
+++ b/django_extensions/management/shells.py
@@ -9,64 +9,90 @@ class ObjectImportError(Exception):
     pass
 
 
+def check_alias(directive):
+    result = directive.split(' as ')
+    if len(result) == 2:
+        object_name, alias = result
+    else:
+        object_name = alias = result[0]
+    return object_name, alias
+
+
 def import_items(import_directives, style, quiet_load=False):
     """
     Import the items in import_directives and return a list of the imported items
 
     Each item in import_directives should be one of the following forms
-        * a tuple like ('module.submodule', ('classname1', 'classname2')), which indicates a 'from module.submodule import classname1, classname2'
-        * a tuple like ('module.submodule', 'classname1'), which indicates a 'from module.submodule import classname1'
-        * a tuple like ('module.submodule', '*'), which indicates a 'from module.submodule import *'
-        * a simple 'module.submodule' which indicates 'import module.submodule'.
+        * a tuple like ('module.submodule', ('classname1', 'classname2 as alias')), which indicates a 'from module.submodule import classname1, classname2 as alias'
+        * a tuple like ('module.submodule', 'classname1[ as alias]'), which indicates 'from module.submodule import classname1[ as alias]'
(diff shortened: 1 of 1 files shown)


django_extensions/management/shells.py
```diff
@@ -9,64 +9,90 @@ class ObjectImportError(Exception):
...
     """
     imported_objects = {}
     for directive in import_directives:
         try:
             # First try a straight import
             if isinstance(directive, six.string_types):
-                imported_object = __import__(directive)
-                imported_objects[directive.split('.')[0]] = imported_object
+                object_name, alias = check_alias(directive)
+                if object_name == alias:
```
trbs: why not have "check_alias" return None for alias when no alias is needed ? then this if statement could just be "if alias"

trbs:
Like this :)

(Now I'm also thinking why we cannot use normal import strings which we parse instead of the complicated list of tuples... so for users the normal python syntax would work...)

django_extensions/management/shells.py
MattFisher: I wanted to avoid having to do 
```
if alias: 
    imported_objects[alias] = imported_object
else:
    imported_objects[object_name] = imported_object
```
after each call to `check_alias`, though I think the "check_alias" method name could be better. Maybe using "name" as the variable name instead of "alias" would be more conceptually clear?

trbs: In https://docs.python.org/3/reference/grammar.html it's called: ```import_as_name``` so we could use ```check_import_as_name``` as the function name which is very "descriptive" :)

trbs:
Could you write some tests for you patch ?
And update the ```docs/shell_plus.rst``` documentation file.

Alternatively if you want I'm thinking of something like:

```python
import ast
import six


def parse_import_line(import_line):
    if not isinstance(import_line, six.string_types):
        raise Exception("Argument must be a string type")
    node = ast.parse(import_line)
    if len(node.body) != 1:
        raise Exception("Invalid import line. Please only specify exactly one import.")
    node = node.body[0]
    if not isinstance(node, (ast.Import, ast.ImportFrom)):
        raise Exception("Only specify import statements")
    from_module = getattr(node, 'module', None)
    name = node.names[0].name
    as_name = node.names[0].asname

    return (from_module, name, as_name)


print parse_import_line("import os")
>>> (None, 'os', None)

print parse_import_line("import os.path")
>>> (None, 'os.path', None)

print parse_import_line("import os.path as ospath")
>>> (None, 'os.path', 'ospath')

print parse_import_line("from os.path import isfile")
>>> ('os.path', 'isfile', None)

print parse_import_line("from os.path import isfile as ospathisfile")
>>> ('os.path', 'isfile', 'ospathisfile')
```

as an alternative way of describing the imports.

People could then just use normal python import lines. Which is far more intuitive... It would (after deprecating the old way) also clear up the code for this.

felocru:
Hello, can you clarify if it is now possible to use aliases? Thank you very much.

trbs:
fixed in master

Issue #579, opened 2014-11-12
Allow override of timestamps when using TimeStampedModel


ses4j:
I have found myself wanting to override the default behavior of TimeStampedModel a few times to manually set the created or modified values.  For instance, when migrating data to a new schema, I want to preserve the timestamps.  Is this possible?

jacobg:
+1

I have a internally used field in a particular model that gets updated frequently, and I don't want to it to change the modified value. I'd like to be able to exclude that field from updating the timestamp, or at least to be able to specify when I save that I don't want the timestamp to be updated.

trbs:
It's currently not possible.

If somebody could write a clean patch for this I'm happy to merge it.

jacobg:
The following is a monkey patch I am using, so you can check out the concept:

```python
def make_timestamp_model_optional():
    from django_extensions.db import fields, models

    #
    # patch field
    #
    orig_pre_save = fields.ModificationDateTimeField.pre_save

    def pre_save(self, model, add):

        if model.no_update_modified:
            return model.modified
        else:
            return orig_pre_save(self, model, add)

    fields.ModificationDateTimeField.pre_save = pre_save

    #
    # patch model
    #
    def save(self, **kwargs):

        self.no_update_modified = kwargs.pop('no_update_modified', False)

        super(models.TimeStampedModel, self).save(**kwargs)

    models.TimeStampedModel.save = save
```

trbs:
@jacobg does this also work for the migration case ? and could you make a pull request for this so we can try and merge it ?

thanks

jacobg:
What would be the concern for migration case? ... I personally don't use South migrations. My app is on Google App Engine, and its nosql datastore does not support it.

trbs:
@ses4j could you comment on this ?

@jacobg Could you make a PR with your patch for including into TimeStampedModel ? I can then always work with @ses4j to make sure it also works inside of migrations.

snoepkast:
@trbs: is this already fixed by any chance? Otherwise I might be able to create a PR, the code posted by @jacobg looks simple enough..

trbs:
Please re-open (or create a new issue) if the PR from @snoepkast didn't completely fix the issue for you.

PR #869, opened 2016-04-12
Add option to preserve modified timestamps on save

snoepkast:
I've added the option to override the default behaviour of always updating the modified date of the `TimeStampedModel` / `ModificationDateTimeField`. I used the code from @jacobg in issue #579 as inspiration.

Also added a simple testcase and some docs.


trbs:
Thanks !

PR #871, opened 2016-04-15
Close Sessions Before Database Drop

brenouchoa:
If there are open connections they need to be closed before the drop of database.
caveat: pg_terminate_backend is only introduced in Postgres 8.4


trbs:
Can we update this so that terminating the backends is optional via a command line flag ?

Issue #866, opened 2016-04-06
graph_models: `exclude_models` with wildcards


decibyte:
I'm using [django-audit-log](https://github.com/Atomidata/django-audit-log) for most of my models. The generated chart from `graph_models` includes all these audit log models. That is not wrong at all, but I'd like to keep them out of my charts, as they are just ruining the overview. Adding the names of all of them (and keeping this list up to date) will be quite tedious.

So, I'd suggest to support wildcards in the `exclude_models` argument, which would allow me to specify `*AuditLogEntry`, to get rid of all models with names that end in `AuditLogEntry`.

I'd be happy to code a solution for this myself and make a pull request for it, but I'd rather not spend the time on it if you have any good reasons to reject this already now.

PR #874, opened 2016-04-19
Graph model filter wildcards

decibyte:
This is a fix to issue #866, which I opened a couple of weeks ago. The basic idea is to allow wildcards in the `include-models` and `exclude-models` arguments to the `graph_models` management command.

Example: `./manage.py graph_models --exclude-models="*IrrelevantModel"` -- this will exclude both `SomeIrrelevantModel` and `OtherIrrelevantModel`.

I've done a bit of restructuring of the code, to allow testing of the function that checks whether a model name should be included or not. I also added tests for the new wildcard code.


trbs:
cool thanks !

PR #877, opened 2016-04-22
Use force_text instead of force_bytes.

DanLipsitt:
This seems to fix some unicode glitches.


trbs:
Could you please include some tests for those glitches ?
(So we have something to test regression on this)

django_extensions/management/modelviz.py
```diff
@@ -210,7 +211,7 @@ def add_relation(field, extras=""):
                     related_query_name = field.related_query_name()
                     if verbose_names and related_query_name.islower():
                         related_query_name = related_query_name.replace('_', ' ').capitalize()
-                    label = '{} ({})'.format(label, force_bytes(related_query_name))
+                    label = '{} ({})'.format(label, force_text(related_query_name))
```
vstoykov: probably this needs to become:
```
label = u'{} ({})'.format(label, force_text(related_query_name))
```
or
```
label = force_text('{} ({})').format(label, force_text(related_query_name))
```

or probably in the beginning of the file to use `from __future__ import unicode_literals`

trbs:
@DanLipsitt are you still interested in updating the PR ?
if not, @vstoykov would you mind creating a new PR so we can fix any issues and merge fixed PR ?

DanLipsitt:
Unfortunately I don't remember what the glitches were off the top of my head but I'll take a look later today.

django_extensions/management/modelviz.py
DanLipsitt: @vstoykov, any thoughts on which format is preferred?

vstoykov: To look more like Python3 probably `from __future__ import unicode_literals` should be put.
I have one concern about this for Python2 and for that reason I think that better approach is `u'{} ({})'` to work best on Python2 and Python3

trbs:
@DanLipsitt thanks !

DanLipsitt:
I updated the branch with a test and a new version of the fix. They are in two separate commits so you can check out the failing test without the fix if you want. Let me know if you want me to squash.

I decided to go with `from __future__ import unicode_literals` because that is consistent with [the Django Project's guidelines](https://docs.djangoproject.com/en/1.10/topics/python3/#syntax-requirements). This carries some risk by affecting other strings in modelviz.py, but tests are passing and I decided it's the simplest way to go forward.

To be honest, it's been a year since I originally made this PR, and I don't remember what the original failing case was. But I think the changes I've made here represent an improvement regardless.

django_extensions/management/modelviz.py
```diff
@@ -8,6 +8,8 @@
   Adapted to be used with django-extensions
 """
 
+# from __future__ import unicode_literals
```
vstoykov: Actually this is commented and will not be executed. Probably it is not needed at all and everything will work regardless.

jrd:
This PR seems reasonable. Why not merge it?

vstoykov:
It should be rebased first probably.

trbs:
Solved the conflict and merged the PR :-)

I would like to ask everybody to check with mainline version to see if it works as expected.

Also if anybody has some time to add additional tests that would be much appreciated.

django_extensions/management/modelviz.py
```diff
@@ -148,7 +151,7 @@ def add_relation(self, field, model, extras=""):
             related_query_name = field.related_query_name()
             if self.verbose_names and related_query_name.islower():
                 related_query_name = related_query_name.replace('_', ' ').capitalize()
-            label = '{} ({})'.format(label, force_str(related_query_name))
+            label = '{} ({})'.format(label, force_text(related_query_name))
```
vstoykov: This line is breaking Python 2.
In the beginning of the file there is commented line `from __future__ import unicode_literals`. Then in Python 2 normal strings are not unicode. When non unicode string is formatted with `format` and there are unicode string as arguments it will rasie `UnicodeEncodeError`.

trbs: Could you maybe make a PR with a test for this ?

vstoykov: I'm tried to reproduce it on my local machine (because it was just guessing based on reading the code) and I found that it can happen only if all of the following are met:

- running on Python 2
- there are `verbose_name` of ForeignKey/ManyToManyFIeld set which contains non ASCII characters
- `graph_models` is executed with `--verbose-names` argument

Did you have any suggestions about the test case what exactly to test?

vstoykov: I created #1247 with test and fix for the issue. You can look at it.

Issue #881, opened 2016-05-03
ValueError: not enough values to unpack


lwm:
So, I installed `django-extensions` from Github with:

```
pip install -e git+https://github.com/django-extensions/django-extensions.git@4e9989efac51f562b6ec9827095cd4012d2b0ebf#egg=django-extensions
```

Then got the following:

```
$ ./manage.py graph_models -a
INFO:root:settings.ENVIRONMENT: dev
Traceback (most recent call last):
  File "./manage.py", line 8, in <module>
    execute_from_command_line(sys.argv)
  File "/home/lwm/work/lock8/velodrome/.venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/lwm/work/lock8/velodrome/.venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/lwm/work/lock8/velodrome/.venv/lib/python3.5/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/lwm/work/lock8/velodrome/.venv/lib/python3.5/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/home/lwm/work/lock8/velodrome/.venv/src/django-extensions/django_extensions/management/utils.py", line 58, in inner
    ret = func(self, *args, **kwargs)
  File "/home/lwm/work/lock8/velodrome/.venv/src/django-extensions/django_extensions/management/commands/graph_models.py", line 147, in handle
    graph_data = generate_graph_data(args, cli_options=cli_options, **options)
  File "/home/lwm/work/lock8/velodrome/.venv/src/django-extensions/django_extensions/management/modelviz.py", line 269, in generate_graph_data
    add_relation(field, '[arrowhead=none, arrowtail=none, dir=both]')
  File "/home/lwm/work/lock8/velodrome/.venv/src/django-extensions/django_extensions/management/modelviz.py", line 241, in add_relation
    target_model = get_model(field.rel.to)
  File "/home/lwm/work/lock8/velodrome/.venv/src/django-extensions/django_extensions/compat.py", line 208, in get_model
    return apps.get_model(*path.split('.', 1))
[...]

Issue #862, opened 2016-03-27
graph_models broken in (at least) django 1.5.1


fabiansvara:
This happens in Django 1.5.1. I have not investigated whether this happens in any other django version.

In ```compat.py```, ```get_models_compat(app_label)``` always returns an empty list, because Django's AppCache instance's app_store SortedDict (in ```db/models/loading.py```) does not contain as keys the app names (which is how the code in compat.py attempts to look up the models), but their model modules.

In ```compat.py```, line 154 etc, changing
```
except ImportError:
        from django.db.models import get_models
        return get_models(app_label)
```
to 
```
except ImportError:
        from django.db.models import get_models
        return get_models(get_app(app_label))
```
fixes it.

The same issue may be present in multiple other locations.

Guess we're all yearning for static type checks.

blueyed:
Maybe `get_models_for_app` should be used here, which uses that pattern?
https://github.com/django-extensions/django-extensions/blob/b3206a7b3179f4b5cfc4d0fe7d8b8638ce25c261/django_extensions/compat.py#L159

blueyed:
Can you try https://github.com/blueyed/django-extensions/commit/dba191e04ee1f64452a8e018c29536f9622c16de / #882, please?

PR #882, opened 2016-05-03
modelviz: look up missing app_label for rel.to field and compat fixes

blueyed:
Fixes https://github.com/django-extensions/django-extensions/issues/881.

Includes a fix for #862, too.


trbs:
Thanks !

PR #883, opened 2016-05-03
compat: move code from try to else block for Django 1.7+

blueyed:
Otherwise this might shadow non-related ImportErrors.


trbs:
Thanks !

PR #884, opened 2016-05-03
Remove and fix usage of `get_models_compat`

blueyed:
It does not expect an app label, but an app instance.

Ref: https://github.com/django-extensions/django-extensions/issues/862.

Trivially conflicts with https://github.com/django-extensions/django-extensions/pull/883.

### DIFF
--- a/django_extensions/compat.py
+++ b/django_extensions/compat.py
@@ -153,20 +153,8 @@ def get_model_compat(app_label, model_name):
         return apps.get_model(app_label, model_name)
 
 
-def get_models_compat(app_label):
-    """Get models on multiple Django versions."""
-    try:
-        # django >= 1.7
-        from django.apps import apps
-    except ImportError:
-        from django.db.models import get_models
-        return get_models(app_label)
-    else:
-        return apps.get_app_config(app_label).get_models()
-
-
 def get_models_for_app(app_label):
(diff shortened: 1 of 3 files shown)


django_extensions/compat.py
```diff
@@ -145,19 +145,19 @@ def get_model_compat(app_label, model_name):
...
 
-def get_models_compat(app_label):
-    """Get models on multiple Django versions."""
+def get_models_compat(app):
+    """Get models on multiple Django versions for an app."""
     try:
         # django >= 1.7
         from django.apps import apps
-        return apps.get_app_config(app_label).get_models()
+        return apps.get_app_config(app).get_models()
```
trbs: https://github.com/django/django/blob/master/django/apps/registry.py#L140

Django calls it app_label here as well and it seems that that function won't work with anything but a string (aka app_label ?).

blueyed: Good point.
Seems you have tried to address this in 3a0ca9e.
But the main problem still is that it behaves differently for Django 1.7 and before.
For Django 1.7 it should be `app.get_models()` then probably.

I've pushed a new commit, which removes it completely and should fix two more commands: 99ecba5.

trbs: Yeah, those were duplicates anyways.
Thanks !

PR #887, opened 2016-05-06
Add a setting to prefix all models in an application

jsferrei:
```SHELL_PLUS_MODEL_ALIASES``` can be kind of tedious for larger applications


trbs:
Thanks !

PR #894, opened 2016-05-16
Document funny graph_models labels for reverse relation names in Python 3

crccheck:
There have been a couple attempts to get the `graph_models` command to behave in Python 3: #851 #834 #877

#851 fixed a bug but created a visual bug where `"groups (user)` now shows as `groups (b'user')`
Using #877 instead of #834 looks like it fixes things, but without tests, it's hard to know what's going on.

So this just adds a basic test. I couldn't think of a way to say "expected failure, but only for Python 3", so I put some skips and threw `FIXME` and `DELETEME` red flags in.

This test was able to catch the problem #851 fixed, but I have no idea what #834 was trying to fix.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -126,7 +126,16 @@ additions for Django projects. See the project page for more information:
     cmdclass=cmdclasses,
     package_data=package_data,
     install_requires=['six>=1.2'],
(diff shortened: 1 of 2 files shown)


trbs:
Thanks

PR #896, opened 2016-05-18
Use unminified version of jQuery bgiframe library.

fladi:
Minification of assets should usually be left the whatever is configured to
handle static assets in a django deploment. It is also a lot easier to fix
bugs in a unminified source file and it removed code redundancies from the
Debian package.


trbs:
Thanks !

Issue #905, opened 2016-05-31
admin_generator don't work with nested apps


marctc:
Taken from @marshalc (on #784)

> I can confirm same issue.
>
>`manage.py admin_generator nhic.ingest` gets me the error (truncated) of:
>
```
File "/Users/carl/.virtualenvs/oxford-processing/lib/python3.5/site-packages/django/apps/registry.py", line 155, in get_app_config
    raise LookupError(message)
LookupError: No installed app with label 'nhic.ingest'. Did you mean 'ingest'?
```
>Whereas the command manage.py admin_generator ingest gets me the feedback of...
```
This command requires an existing app name as argument
Available apps:
    django.contrib.admin
    django.contrib.auth
    django.contrib.contenttypes
    django.contrib.sessions
    django.contrib.sites
    django_extensions
    nhic.ingest
```
... so you can see the circular logic here suggesting there's a bug. Unfortunately I don't have time presently to investigate and debug this (this is my first time trying to use this extension), but if you want a practice project/model, you can see the current state of what I'm using this with at https://github.com/nhic-tra/oxford-processing/tree/1748b5406b9d17397b235ccd464016dc3327060c

>This is under:
>
>   * Django (1.9.6)
>   * django-extensions (1.6.8) <-- was 1.6.7, tested and reconfirmed after update)
>   * Python 3.5.1

PR #906, opened 2016-05-31
Fix bug #905 in admin_generator command

marctc:
Replace `get_models_for_app` for `get_models_from_cache` when getting models for a
given app.


trbs:
Thanks !

PR #915, opened 2016-06-25
Remove support for end-of-life Python 3.2.

timgraham:
Django 1.8 will officially drop support for it at the end of the year, but most apps I've seen have already done so and anyone still using it can use an older version of django-extensions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
   - TOX_ENV=py27-dj19
   - TOX_ENV=py27-dj110
   - TOX_ENV=py27-djmaster
-  - TOX_ENV=py32-dj18
   - TOX_ENV=py33-dj18
   - TOX_ENV=py34-dj18
   - TOX_ENV=py34-dj19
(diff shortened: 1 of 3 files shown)


trbs:
I would like to follow the Django policy on this: https://docs.djangoproject.com/en/1.9/faq/install/#what-python-version-can-i-use-with-django

So we can leave this open and by the end of the year merge the PR to remove Py32.

trbs:
we should also remove the "virtualenv<14.0.0" from .travis.yml once we drop python 3.2 support

timgraham:
@trbs, can we merge this now that the Python 3.2 build is an expected failure (looks like pytest dropped compatibility)? I don't see much value in try to fix it at this point. I doubt Python 3.2 users need the latest and greatest django-extensions anyway.

trbs:
Yeah, py.test removing 3.2 support forced our hands here a little bit.
I want to remain true to our policy of following Django's supported Python versions as much as possible but in this case I'm okay with making the exception.

People should not be using Python 3.2 (please upgrade to Python 3.5 or higher if you can) or Django 1.8 (please upgrade to 1.10 or higher).

trbs:
Thanks @timgraham !

PR #920, opened 2016-06-26
Make mention of iPython Notebook easier to find

ipsod:
The mention of this important feature was buried in the middle of an arcane section on configuration.

Also added a line that points out two of iPython Notebooks most useful features for Django developers - the ability to save the notebook, and the ability to reload the kernel in an open shell.


trbs:
Thanks !

PR #922, opened 2016-06-28
AutoSlugField: fix allow_duplicates

### LINKED ISSUE #801: AutoSlugField doesn't populate model
Using the AutoSlugField in the normal way (making a slug from a name). It seems the slug field doesn't get populated in python but only in the sql.

Create a model and slug field is blank. Load the model from the database and the slug is populated.

Means I can't use the new model to redirect to since the slug is empty.

beshrkayali:
Resolves #801 

The issue is that if `allow_duplicates` is True, `create_slug` will return the slug directly without setting the attribute (field name) on the model instance. `find_unique` takes care of this in case `allow_duplicates` is False or not set.


trbs:
Thanks !

PR #926, opened 2016-07-04
Fixed setting ipaddr:port for runserver_plus for django 1.10

kosmos342:


trbs:
@kosmos342 why did you close the issue ?

kosmos342:
@trbs because it works fine with master branch

trbs:
added your patch in f3e62202b82dbeee348c7b79015e76db33b59564
thanks !

Issue #927, opened 2016-07-05
graph_models error with pydot lastest version


Eric-Bujeque:
I'm working with python 2.7.11+ and latest version of pydot (1.2.2) and pyparsing (2.1.5). When I run:

`django-admin graph_models -o models.png`

I get:

`Traceback (most recent call last):
  File "/home/ebujeque/.virtualenvs/gmp/bin/django-admin", line 11, in <module>
    sys.exit(execute_from_command_line())
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django_extensions/management/utils.py", line 58, in inner
    ret = func(self, *args, **kwargs)
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django_extensions/management/commands/graph_models.py", line 164, in handle
    self.render_output_pydot(dotdata, **options)
  File "/home/ebujeque/.virtualenvs/gmp/local/lib/python2.7/site-packages/django_extensions/management/commands/graph_models.py", line 221, in render_output_pydot
    graph.write(output_file, format=format)
AttributeError: 'list' object has no attribute 'write'`

I tried with pyparsing(1.5.7) and pydot(1.2.2), but it fails too. I solved it with:
- pyparsing==1.5.7
- pydot=1.1.0

But It can't work with lastest version of both libraries.

x007007007:
me too, I 'm working on python3.5

vhdmsm:
And me too.

trbs:
Please check with the github version which has @x007007007 PR was merged

trbs:
Please reopen / create a new ticket if there still issues.

Jufik:
[...]

PR #931, opened 2016-07-25
fix issue #927

x007007007:
according to http://stackoverflow.com/questions/38176472/graph-write-pdfiris-pdf-attributeerror-list-object-has-no-attribute-writ
try import pydotplus as default pydot


trbs:
Thanks !

PR #934, opened 2016-08-01
Fixes JSONField for Django 1.9+

wsantos:
JSONField was failing with the latest master changes, manly because we were using a mutable object as default and this lead us to a problem where all objects with out default were sharing the same mutable object.

I've also included more tests that cover my usage scenarios.


trbs:
Thanks !

PR #946, opened 2016-08-26
Fixed a bug in joining urls. urllib is the way to go to join urls.

Cediddi:
I found this while I was contributing to https://github.com/akademikbilisim/ab-kurs-kayit
The bug was causing an url like this: `/admin/userprofile/userprofile//foreignkey_autocomplete/` (both app and model name are the same, problem is the double slashes.)

### DIFF
--- a/django_extensions/admin/widgets.py
+++ b/django_extensions/admin/widgets.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import six
+from six.moves import urllib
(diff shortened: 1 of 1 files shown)


trbs:
Thanks !

Cediddi:
Thank you for being extremely fast (13 minutes!) :heart:

trbs:
new release should be out shortly :)

PR #948, opened 2016-08-28
Fix reading templates setting after django 1.8

### LINKED ISSUE #942: validate_templates isn't valid when migrated to the 1.8 TEMPLATES setting.
`TEMPLATE_DIRS` is deprecated in favor of `TEMPLATES['DIRS']`. If the setting doesn't exist this will fail:

https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/validate_templates.py#L54

albertyw:
This replaces code that directly reads `TEMPLATE_DIRS` with a function that can read either that setting or the newer `TEMPLATES` setting.

Fixes #942

### DIFF
--- a/django_extensions/compat.py
+++ b/django_extensions/compat.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 import importlib
 import django
 
+from django.conf import settings
+
 
 #
 # Django compatibility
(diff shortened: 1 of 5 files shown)


trbs:
Could you please move the code for reading pre/post Django 1.8 TEMPLATE settings to ```django_extensions/compat.py``` ?

trbs:
Thanks !

PR #953, opened 2016-09-04
Use new location for url reverse import

michael-k:
The `reverse` function is located in `django.urls` since Django 1.10, see https://docs.djangoproject.com/en/1.10/ref/urlresolvers/#module-django.urls.

Although the old imports "will continue to work until Django 2.0" we should not encourage users to use the old location for new import statements.


trbs:
Thanks !

PR #957, opened 2016-09-06
consistent badge design - using .svg

gesellc:
Sorry for the noise. Didn't catch this with django-extensions/django-extensions#956.


trbs:
No problem, thanks ! all help is appreciated 👍

PR #962, opened 2016-09-12
Fix argument error when running dumpscript

steph-ben:
Without this fix, the dumpscript command fail:

```
manage.py dumpscript auth.User
[...]
  File "/usr/lib/python2.7/site-packages/django_extensions-1.7.4-py2.7.egg/django_extensions/management/commands/dumpscript.py", line 95, in handle
    app_labels = options['appname']
KeyError: 'appname'
```


trbs:
We should catch this in our tests. So we can also see which versions of Django are effected.

For this specific issue can't we not fix this in ```add_arguments``` ?

Issue #955, opened 2016-09-05
More sensible error message for runscript


Deleetdk:
I wasted a lot of time discovering that runscript gives the same error message for when there is an import error in the script and when it cannot find the script at all. Please add a check for whether the file exists at all first, and throw a distinct error if it cannot find any such script.

[StackOverflow example.](http://stackoverflow.com/questions/39322967/djangos-runscript-no-valid-module-for-script-filename-found)

kevgathuku:
Hey @Deleetdk, I have just been looking through this and I would like to tackle it. Sorry for the bad experience you had to go through to discover this bug.

After looking into it, I think the error message for a non-existent script is correct. 
i.e. `No (valid) module for script 'test_script' found`

I think the error message for when there's an import error in the script is what needs to be fixed.
e.g. In the case where `import nonexistrentpackage` is at the top of the file.

However, you will notice that when the invalid import is placed inside the `run` function, the correct error message is raised. For example, see the following code and the error message it raises.
```python
def run():
    import nonexistent
    print("Test")
```

Exception:
```
...
ImportError: No module named nonexistent
```

I think this is what should be raised when there is an invalid import. 
What do you think?

Deleetdk:
Makes sense. However, I didn't know that I had to put my imports inside `run()` too. I noticed that [the official docs](http://django-extensions.readthedocs.io/en/latest/runscript.html) also have imports outside `run()`. It's not a common coding pattern to do, so it's likely that many others will run into this same problem.

kevgathuku:
@Deleetdk that's not the recommended way to do it. I just tried it out and noticed it handled the error correctly.
I'll try to see whether I can replicate that kind of error even when the invalid import is at the top level.

trbs:
The imports inside the ```run()``` function should not be necessary. It is a workaround to avoid some of the issues with unclear messages.

Thanks to @kevgathuku we have #964 now !
I would like to see us fixing all potential problems with errors in the scripts not just ```ImportError```. We have a similar issue with ```AttributeError``` as well.

trbs:
Would be cool if we could get test cases as well for this to avoid regressions in the future.

hhstore:
```
No (valid) module for script 'test_script' found
```

## Reason:
- sometimes, this error was  caused by `your codes have bugs`.
- like: 
    - your code `import a third party package`, and `the package has bugs`, 
    - It is difficult for you to find errors `caused by  the third party package`.


## How to fix it:
[...]

PR #964, opened 2016-09-18
Display correct error message for missing imports

kevgathuku:
When a missing import is found in a script, this changes the error
message displayed to indicate that there's a missing import, rather than
giving a generic error message similar to the one raised when the script
does not exist

For a case such as this:
`test_script.py`

``` python
import nonexistrentpackage

def run():
    print("Test")
```

The error message raised previously was:

```
No (valid) module for script 'test_script' found
Try running with a higher verbosity level like: -v2 or -v3
```

This changes the error to be more descriptive, without requiring the user to change the default verbosity mode, i.e.

```
Cannot import module 'scripts.test_script': No module named nonexistent.
No (valid) module for script 'test_script' found
Try running with a higher verbosity level like: -v2 or -v3
```


kevgathuku:
Hey @trbs don't mean to be a bother, but have you had a chance to look at this patch? It fixes issue #955
Is there anything else I could do to make it better?

trbs:
@kevgathuku no problem. I was traveling over the last few days so I did not get the time to look at PR's.

Not really sure why AttributeError and ImportError where caught here originally. I think we can split it up in two ```except``` sections. As I think now an AttributeError in the script will show the message as if it where an ImportError.

Also thinking; why not just let the import script throw it's original errors ? and get rid of the try-except altogether ?

The purpose of the try-except statement here only seems to be to catch if the script exists or not.. 
For that we could also use something like ```importlib.util.find_spec(mod)```, ```pkgutil.find_loader(mod)``` or ```imp.find_module(mod)```.

Wouldn't that be even better ?

kevgathuku:
Hi @trbs,
Thanks for the reply. I would need to look into those options first, in order to understand the options we could use.

kevgathuku:
Hey @trbs seems I'm working on this now again. Apologies for the extremely long wait.
It seems the `AttributeError` was introduced in this commit: https://github.com/django-extensions/django-extensions/commit/fc6722dac01119d5c26125330a5a741c9f644869

I think the key thing now is differentiating when a script raises an import error vs when a script is not found. 
Right now, a valid script will not be added to [`modules`](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/runscript.py#L156) if it raises an `ImportError`. 
Making that distinction is what I'm working on right now. Expect an update on that soon.

kevgathuku:
I've opted to raise the error if an invalid import is found in one of the scripts. Let me know what you think of this approach.

kevgathuku:
On the other hand, I am having some trouble testing the invalid imports. I keep getting this error, even in the collect tests phase. Any idea on how to prevent the script from being run before the tests run?
image

trbs:
Very late reply 🥇 :-)

I think this is because `py.test` is searching for docstrings, it scans through all the files to do that. You can exclude specific files in `setup.cfg` so that it will not error out on  `invalid_import_script.py`.

trbs:
Thanks !

kevgathuku:
Hey @trbs thanks for merging, but there's still some pending issues that I noticed I didn't fix 😅 
This is mostly caused by `importlib.import_module` behaving differently after Python 3.3.
I'll fix that and open another PR with some more tests 😄

trbs:
Thanks a lot !

PR #967, opened 2016-09-27
Fixed `app_name` resolving in shell_plus

nazarewk:
`app_name` was improperly resolved in case `models.py` was split into multiple files.
# Example

``` python
from myapp.models.racing import Race
```
## Old behaviour

``` python
app_name == 'models'
```
## Expected (and fixed) behaviour

``` python
app_name == 'myapp'
```

### DIFF
--- a/django_extensions/management/shells.py
+++ b/django_extensions/management/shells.py
@@ -108,6 +108,7 @@ def import_items(import_directives, style, quiet_load=False):
 
 def import_objects(options, style):
     from django.apps import apps
+    from django.apps.config import MODELS_MODULE_NAME
     from django import setup
 
     if not apps.ready:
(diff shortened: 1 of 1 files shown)


trbs:
Looks good.
Could you please add tests for the get_app_name function and tests for regressions in ```shell_plus``` ?

domandinho:
Looks good hovewer we can retrieve app name from meta of model. If we have only model path it is not a problem  because we can import this model from this string.
See django-extensions/django_extensions/collision_resolvers.py:20
```
    @classmethod
    def get_app_name_and_model(cls, full_model_path):
        model_class = import_string(full_model_path)
        return model_class._meta.app_config.name, model_class.__name__
```

trbs:
Could anybody also look at creating tests for this ? Otherwise I will sadly have to close this PR.

nazarewk:
I've rebased and written simple doctests, but i don't have time to make full test setup of the issue, theoretically doctest should be enough

@domandinho Tests are not passing when i give it full `app_config.name`, extensions expect only last part, also in most places the model isn't present yet.

trbs:
Thanks for the update !

django_extensions/management/shells.py
```diff
@@ -34,6 +36,29 @@ class ObjectImportError(Exception):
...
+    'testapp'
+
+    'testapp' instead of 'some.testapp' for compatibility:
+    >>> get_app_name('some.testapp.models.foo')
+    'testapp'
+    """
+    parts = mod_name.split('.')
+    try:
+        try:
+            return parts[parts.index(MODELS_MODULE_NAME) - 1]
```
trbs: Should we use:
```python
rparts = list(reversed(parts))
return rparts[rparts.index(MODELS_MODULE_NAME) + 1]
```

for something like:
```python
>>> get_app_name('some.models.testapp.models.foo')
'testapp'
```

Avoiding getting `some` as a result there.

nazarewk: good catch

trbs:
Thanks !

PR #971, opened 2016-10-11
Rewrite modelviz module from method to class based processor.

GrivIN:
This change will make changes to models graph easier.

For ex. to add model docstring into graph it is not needed to copy whole module any more.


trbs:
Many Thanks !

May I ask you to look at the tests as well ? and maybe write some additional ones for better coverage.

GrivIN:
sure, I will take a look, see you soon! ;)

PR #973, opened 2016-10-14
Fix that aliases from SHELL_PLUS_MODEL_ALIASES were not applied.

etienned:
L141 removal is just a cleanup, `app_aliases` was not used in this scope.

Aliases were never applied because `alias` was always reset to `model_name` when they were no prefix on L176.


trbs:
Thanks.

Could you maybe write some tests for this as well ?

PR #978, opened 2016-10-24
Track stable versions of pip requirements

federicobond:
This patch changes the calculation of the latest available version in `pipchecker`:

 **Before**
pipchecker took the latest visible release.

```
celery 3.1.23                    4.0.0rc6 available
```

**Now**
pipchecker will search for all releases (visible and hidden) and select the latest stable one if your declared version is stable (or the latest unstable if not).

```
celery 3.1.23                    3.1.24 available
```

```
celery 4.0.0rc3                  4.0.0rc6 available
```

I would love to hear your feedback about this. If you are wary about changing previous behaviour but think this is a good idea, I could add a command line flag.


trbs:
Thanks !

PR #979, opened 2016-10-31
Django updated the admin search raw id HTML to remove the image

streeter:
The search image [is no longer included](https://github.com/django/django/blob/master/django/contrib/admin/widgets.py#L174-L177) in the HTML for the widget html.


trbs:
Which versions of Django are we talking about ?
Does this work for all versions of Django that are currently supported by Django-Extensions ?

jrd:
Seems to not be there since Django 1.11 (the current minimum supported version)

trbs:
Cool thanks, so then it should be safe to merge the PR :-)

PR #980, opened 2016-11-03
Add autocomplete for inline model admins. Fixes #109

### LINKED ISSUE #109: Use the ForeignKeyAutocompleteAdmin form as in a InlineModelAdmin
Hi,

It seems that there is no way to use the ForeignKeyAutocompleteAdmin form as in a InlineModelAdmin form.

I wanted to display a foreign key in a TabularInline form in my admin, but it seems impossible. That would be an awesome feature to add.

Regards,
Thibault

wolever:
It turns out that the existing admin autocomplete code works perfectly with inline admins; the base class just needs to be refactored into a mixin and a couple little JavaScript things fixed.

### DIFF
--- a/django_extensions/admin/__init__.py
+++ b/django_extensions/admin/__init__.py
@@ -15,12 +15,12 @@ from django.db.models.query import QuerySet
 from django.utils.encoding import smart_str
 from django.utils.translation import ugettext as _
 from django.utils.text import get_text_list
-from django.contrib.admin import ModelAdmin
+from django.contrib import admin
 
 from django_extensions.admin.widgets import ForeignKeySearchInput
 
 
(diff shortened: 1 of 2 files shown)


django_extensions/admin/__init__.py
```diff
@@ -15,12 +15,12 @@
 from django.utils.encoding import smart_str
 from django.utils.translation import ugettext as _
 from django.utils.text import get_text_list
-from django.contrib.admin import ModelAdmin
+from django.contribimport import admin
```
camilonova: This break tests https://travis-ci.org/django-extensions/django-extensions/builds/173107246

Maybe is `from django.contrib import admin`

wolever:
D'oh! Apparently I'm not great at copy+pasting from my codebase.

Pushed an update which I've also run tox over.

PR #983, opened 2016-11-14
Added nopin option to disable pin in runserver_plus

deese:
Using the last source code from github.
Sorry for my fights with github :)

As it uses an environment variable, if the version doesn't support pin, the variable is not used.


trbs:
Thanks !

PR #990, opened 2016-12-01
Add django.contrib.auth.get_user_model to shell+

piotr-szpetkowski:
Sometimes `get_user_model` is much more convenient than fetching the model class by name.

There are some use cases, where people could find it useful. I for example use it to fetch swapped `User` model and to find out which one is currently active and writing `from django.contrib.auth import get_user_model` is quite cumbersome.

Also, it makes it easier to approach a new project, especially if it overrides `User` model in unconventional way.


trbs:
thanks !

PR #995, opened 2017-01-03
Allow clear_cache to specify which cache is to be cleared

goodtune:
- Existing command line interface is maintained
- Providing the optional --cache argument you can alter behaviour
- Output now reflects the cache key that was cleared


trbs:
Thanks !

Issue #998, opened 2017-01-15
error when generating graph with either include or exclude set


tshabs:
In python 2:

`anagement/modelviz.py", line 95, in generate_graph_data
    if relation['target'] in nodes:
TypeError: 'NoneType' object has no attribute '__getitem__'
`

In python 3:

`nt/modelviz.py", line 95, in generate_graph_data
    if relation['target'] in nodes:
TypeError: 'NoneType' object is not subscriptable`

For now I have a fix by adding an if in line 95 to check if relation is None (PR to follow).

GabLeRoux:
Thanks for this, I confirm it works, will wait for `django-extensions==1.7.6`.

I needed `graph_model` quickly so I did this and then it worked;
```bash
pip uninstall django-extensions
pip install git+https://github.com/django-extensions/django-extensions#egg=django-extensions
```

trbs:
New release should be out shortly

PR #999, opened 2017-01-15
fix for issue #998 error when generating graph with either include or…

tshabs:
… exclude set

please see https://github.com/django-extensions/django-extensions/issues/998


trbs:
Thanks !

PR #1002, opened 2017-01-25
Fixed #1001 - Use django.urls when available.

### LINKED ISSUE #1001: Prefer django.urls instead of django.core.urlresolvers
`django.core.urlresolvers` is deprecated and emits warnings.

pelme:


trbs:
Thanks !

PR #1003, opened 2017-01-25
Improved displaying of related field names in diagram

lockie:
Hello!
This little MR drops 'b' prefix and the quotes from the related query names in diagram (otherwise it looks liek this: looks).


trbs:
Thanks !

PR #1007, opened 2017-02-02
Fix `attributes` referenced before assignment

maxnordlund:
This fixes the `--hide-relations-from-fields` flag, or at least that is what triggered this bug in the first place.


trbs:
Thanks !

PR #1012, opened 2017-02-12
Quote table name when using PRAGMA table_info

dsanders11:
Table names with hyphens will cause the `sqldiff` management command to fail when using sqlite3. Creates the following error:

````django.db.utils.OperationalError: near "-": syntax error````

Fix is simply to quote the table name. Issue may exist for other database types as well (such as MySQL).


trbs:
Thanks !

PR #1027, opened 2017-03-08
Added decorator to modify fields

tillkolter:
It happened to me a few times now, that I needed to overwrite certain field attributes defined in an abstract parent model. I found the actual code snipped on stackoverflow, but thought it this was a nice little decorator which other people could need as well.

### DIFF
--- a/django_extensions/db/decorators.py
+++ b/django_extensions/db/decorators.py
@@ -1,0 +1,48 @@
+
+
+def modify_fields(**kwargs):
+    """
+        A decorator to modify field properties for a child model
+        With regards to http://stackoverflow.com/a/24475838/6245268
+
+        Usage:
+            ```
+                class AbstractVehicle(models.Model):
+                    owner = models.ForeignKey(User)
+
+                    class Meta:
+                        abstract = True
+
+
+                @modify_fields(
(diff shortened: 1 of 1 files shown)


trbs:
Thanks for the PR !

I'm a bit on the fence for this one. Not sure if it fits into `django-extensions`, it would also need to be maintained.

If it would be included we would need test cases and updates to documentation as well.

For now I will leave it open, hopefully additional people will come and endorse this PR so we know there is a need in the community for this decorator (and open the door to other decorators).

vstoykov:
From Django 1.10 you can overwrite fields from abstract classes:
- https://docs.djangoproject.com/en/1.10/releases/1.10/#models
- https://docs.djangoproject.com/en/1.10/topics/db/models/#field-name-hiding-is-not-permitted

With this you can also override not only parameters of the field but even the type of the field or even remove it if you don't need it.

I think that this feature is not needed in django-extensions, because it should be deprecated immediately in favor of Django 1.10+ built in functionality.

trbs:
Thanks @vstoykov and @tillkolter for contributing ! We appreciate everybody's help and interest in `django-extensions`.

PR #1030, opened 2017-03-16
Test against Django 1.11

vstoykov:
As of Django 1.11 `MigrationWriter.as_string` returns `unicode` not `bytes` and for that reason I changed the test like that.

More details at that change https://github.com/django/django/commit/979ea95608176e9c88678fd4d8a096c29654b5f3


vstoykov:
Is it possible the build to be retired because tests are passing but for py35-dj110 it says it is failing because sending coverage results to coveralls stalled?

trbs:
I restarted the build and now it went thought :)

Thanks !

PR #1035, opened 2017-03-22
"ip:port" specification for runprofileserver

landswellsong:
I noticed that countrary to what the descriptions say, `runprofileserver` doesn't accept the usual `host:ip` pair of the `runserver` command. I added the relevant bit straight from the latter command's code. Seems to work in my setup.

Please let me know if I'm doing something wrong, looked like a bug for me.

UPD: re-added `--nothreading` too.

### DIFF
--- a/django_extensions/management/commands/runprofileserver.py
+++ b/django_extensions/management/commands/runprofileserver.py
@@ -260,7 +260,7 @@ class Command(BaseCommand):
                     if use_static_handler and (settings.DEBUG or insecure_serving):
                         handler = StaticFilesHandler(handler)
                 handler = make_profiler_handler(handler)
-                run(addr, int(port), handler)
+                run(addr, int(port), handler, threading=options.get('use_threading', True))
             except socket.error as e:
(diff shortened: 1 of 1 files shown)


trbs:
How does the '--nothreading' option work now, shouldn't that be passed on to `run(...)` ?

landswellsong:
@trbs I'm not a big expert on django, but I just assume it should be the way `runserver.py` [handles it](https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L46). It's then taken from the `options` and used in `inner_run()` [here](https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L114) and to `run()` [here](https://github.com/django/django/blob/master/django/core/management/commands/runserver.py#L142).

trbs:
Thanks I will try to get to merging this PR somewhere the next couple of days.

landswellsong:
Thank you very much.

trbs:
Thanks !

Issue #1029, opened 2017-03-14
`setup.py test` broken


DanLipsitt:
`setup.py test` doesn't work without `--pytest-args` arguments:

```
$ python setup.py test
running test
Searching for mock
Best match: mock 2.0.0
Processing mock-2.0.0-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/mock-2.0.0-py2.7.egg
Searching for pytest-cov
Best match: pytest-cov 2.4.0
Processing pytest_cov-2.4.0-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/pytest_cov-2.4.0-py2.7.egg
Searching for pytest-django
Best match: pytest-django 3.1.2
Processing pytest_django-3.1.2-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/pytest_django-3.1.2-py2.7.egg
Searching for python-dateutil
Best match: python-dateutil 2.6.0
Processing python_dateutil-2.6.0-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/python_dateutil-2.6.0-py2.7.egg
Searching for pbr>=0.11
Best match: pbr 2.0.0
Processing pbr-2.0.0-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/pbr-2.0.0-py2.7.egg
Searching for funcsigs>=1
Best match: funcsigs 1.0.2
Processing funcsigs-1.0.2-py2.7.egg

Using /Users/dan/src/django-extensions/.eggs/funcsigs-1.0.2-py2.7.egg
Searching for coverage>=3.7.1
Best match: coverage 4.3.4
Processing coverage-4.3.4-py2.7-macosx-10.8-x86_64.egg

Using /Users/dan/src/django-extensions/.eggs/coverage-4.3.4-py2.7-macosx-10.8-x86_64.egg
running egg_info
writing requirements to django_extensions.egg-info/requires.txt
writing django_extensions.egg-info/PKG-INFO
writing top-level names to django_extensions.egg-info/top_level.txt
writing dependency_links to django_extensions.egg-info/dependency_links.txt
reading manifest file 'django_extensions.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'django_extensions.egg-info/SOURCES.txt'
running build_ext
Traceback (most recent call last):
  File "setup.py", line 155, in <module>
    'Topic :: Utilities',
[...]

PR #1038, opened 2017-04-03
#1029, fix broken setup.py test

rootart:


trbs:
Could we change this so that we unify / make DRY this patch with the Makefile ?

That should also help avoiding regression in the future.

rootart:
@trbs I think keeping only `python setup.py test` should work fine. Thanks.

trbs:
Thanks !

PR #1039, opened 2017-04-05
Add hook to get imported objects

sloria:
Add a hook to `shell_plus.Command` for getting the auto-imported objects. This allows subclasses to add objects at run time.

Example use case:

```python
from django_extensions.management import shell_plus

class Command(shell_plus.Command):
    def get_imported_objects(self, options):
        imported_objects = super().get_imported_objects(options)
        me_username = getattr(settings, 'MY_USERNAME', None)
        if me_username:
            me = User.objects.get(username=me_username)
            imported_objects['me'] = me
        return imported_objects


trbs:
Thanks !

PR #1040, opened 2017-04-05
Treat rc versions as unstable in pipchecker

federicobond:
This prevents pipchecker from suggesting upgrading to a release candidate when the currently installed version is stable.


trbs:
Thanks !

trbs:
p.s. Thinking we might want to add a cli flag to allow `pipchecker` to show non-stable versions ?

federicobond:
I don't personally find it very useful tbh. In general, if I have a dependency on a stable version I want to remain in that stable channel. If the dependency is unstable, then I probably want to update to the latest version, even if it's unstable.

PR #1042, opened 2017-04-07
Fix #1011 AutoSlugField for foreignkey relationships

### LINKED ISSUE #1011: AutoSlugField for foreignkey relationships
Thanks for this great library! I use it on every Django project.

I was looking to use foreignkey relationships in my autoslugfield. Something like:

```
class TeamMember(models.Model):
    name = CharField
    team = FK(Team)
    slug = AutoSlugField(populate_from=("team__name", "name")

class Team(models.model):
    name = CharField
```

Would a PR for this be desirable? I was thinking of using something like this StackOverflow answer to lookup the foreignkey relationship: http://stackoverflow.com/questions/20235807/how-to-get-foreign-key-values-with-getattr-from-models

YPCrumble:
Fixes #1011 

This PR allows the user to span ForeignKey relationships with AutoSlugField's `populate_from` argument. It also allows the user use model methods in the `populate_from` argument. 

@trbs please let me know any feedback or suggestions? Thanks again for your work on this library, and for considering this PR!

### DIFF
--- a/django_extensions/mongodb/fields/__init__.py
+++ b/django_extensions/mongodb/fields/__init__.py
@@ -10,6 +10,7 @@ import re
 import six
 import datetime
 from django import forms
+from django.db.models.constants import LOOKUP_SEP
 from django.template.defaultfilters import slugify
 from django.utils.translation import ugettext_lazy as _
 from mongoengine.fields import StringField, DateTimeField
@@ -95,7 +96,7 @@ class AutoSlugField(SlugField):
 
         if add or self.overwrite:
             # slugify the original field content and set next step to 2
-            slug_for_field = lambda field: self.slugify_func(getattr(model_instance, field))
+            slug_for_field = lambda lookup_value: self.slugify_func(self.get_slug_fields(model_instance, lookup_value))
             slug = self.separator.join(map(slug_for_field, self._populate_from))
             next = 2
         else:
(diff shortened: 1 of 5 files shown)


trbs:
Looks good ! only thing I can think of is if Django (ORM) already has a function for doing the traversal lookup.

YPCrumble:
You can grep for `LOOKUP_SEP` in the docs (this defaults to `__` in django.db.models.constants), which shows you how the ORM uses the query expressions separated by `__`. 

The issue is the ORM needs to search for all kinds of things like `__icontains`, etc., as well as field names. It also appears to be building a raw SQL query with joins directly vs. using `LOOKUP_SEP` to traverse ORM relationships as we are here. I didn't see a utility function that I could easily apply. 

Most of the code is in https://github.com/django/django/blob/master/django/db/models/query.py if you'd like to verify/let me know if I'm missing anything?

django_extensions/mongodb/fields/__init__.py
```diff
@@ -145,6 +145,20 @@ def create_slug(self, model_instance, add):
             next += 1
         return slug
 
+    def get_slug_fields(self, model_instance, lookup_value):
+        lookup_value_path = lookup_value.split('__')
```
trbs: Lets use LOOKUP_SEP instead of '__' ?

trbs:
I also cannot find a `Django` utility function for doing the traversal lookup.

Could imaging there are some edge cases with a straight `getattr(...)` but we can also fix them later.
I added two comments to the PR, could you please take a look at those ?

YPCrumble:
@trbs thanks for the review! I only saw one comment - using `LOOKUP_SEP` instead of `__` which I just added to the commit. Let me know if any other changes?

trbs:
Ah sorry I hit the review button instead of "add comment" for the second question :)
Wondering about the "return None" and if we should have it fail loudly instead of "swallowing" the exception.

YPCrumble:
@trbs agreed - just pushed an updated version. Thanks for the review!

django_extensions/mongodb/fields/__init__.py
```diff
@@ -145,6 +146,20 @@ def create_slug(self, model_instance, add):
...
         return slug
 
+    def get_slug_fields(self, model_instance, lookup_value):
+        lookup_value_path = lookup_value.split(LOOKUP_SEP)
+        attr = model_instance
+        for elem in lookup_value_path:
+            try:
+                attr = getattr(attr, elem)
+            except AttributeError:
+                return None
```
trbs: Did you intentionally kept the `return None` here ?
Thinking might be best to also throw exception in case of Mongo.

YPCrumble:
@trbs I'm sorry!!!!!!! Forgot about the mongodb file. Just updated to raise the error there as well.

trbs:
No worries, thank you for the PR and all the work you put into this !

PR #1043, opened 2017-04-09
Use same algo to generate SECRET_KEY as Django

coagulant:
Using random from standard library is not cryptographically secure.

Given Django has project template support this management command might not be necessary anyomore. It's better delete it or remove insecure code


trbs:
Hi it seems that function (or at least in that form) is only available in later versions of Django.

Maybe the name has changed or please add a try-except around it with the old code for older version of Django.

Thanks !

trbs:
Great thanks !

PR #1050, opened 2017-05-03
Add syntax highlighting when printing SQL

maxnordlund:
This uses the excellent pygments library to to the heavy lifting. In the process of adding this to both the runserver and shell command, the few differences between them are unified.

skarmavbild 2017-05-03 kl 15 40 44


trbs:
Very cool !

For the commit I'm wondering are we missing the print / logger lines for `raw_sql` ?

maxnordlund:
Whops, you're completely right, I must've deleted them by accident. I'll fix that right away.

trbs:
Thanks !

maxnordlund:
No problem 😸

PR #1051, opened 2017-05-05
supported_backends changed.

mumino:
in django 1.9, 'django.db.backends.postgresql_psycopg2' renamed to 'django.db.backends.postgresql'.


trbs:
Thanks !

PR #1052, opened 2017-05-15
Fix for hanging when printing large SQL quires

maxnordlund:
When printing large quires it would hang, this just cuts it off to ensure it doesn't go out of hand. A common large query is bulk insert.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -148,7 +148,8 @@ class Command(BaseCommand):
                         raw_sql = self.db.ops.last_executed_query(self.cursor, sql, params)
 
                         if sqlparse:
-                            raw_sql = sqlparse.format(raw_sql, reindent=True)
+                            raw_sql = raw_sql[:1000]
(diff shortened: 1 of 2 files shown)


trbs:
Can you add a configuration option for the cutoff point ? Then if users want to set it to truncate to shorter (or longer) queries they could do so.

Thanks !

maxnordlund:
Sure thing, what would be a good name you recon?

trbs:
Was thinking something like `settings.EXTENSIONS_PRINT_SQL_TRUNCATE`.

For `shell_plus` there is already `settings.SHELL_PLUS_PRINT_SQL` however there is no `runserver_plus` equivalent of that. 

Maybe `settings.SHELL_PLUS_PRINT_SQL_TRUNCATE` and `settings.RUNSERVER_PLUS_PRINT_SQL_TRUNCATE` ?

But I'm completely open to other suggestions as well :)

maxnordlund:
Sounds like a plan.

trbs:
Thanks !

maxnordlund:
No problemo

Issue #75, opened 2011-02-24
Exclude models


ChrisSwart:
I would like to exclude certain models from the generated file.

I use history tables (see book ProJango by Marty Alchin Chapter 11 (page 263 onwards) and 
http://qr7.com/2010/10/django-simple-history-ftw/ by Corey Bertram). I prefix the models with zzhistorical (so it is listed at the end of the tables in pgAdmin) and would like to either exclude it with a file containing all the model names to exclude or a wildcard format e.g. zzhistorical*

Thank you

PR #1053, opened 2017-05-15
Document graph including/excluding specific models

morenoh149:
Had to dig to find https://github.com/django-extensions/django-extensions/issues/75#issuecomment-823840 here it is added to the docs


trbs:
Thanks !

PR #1058, opened 2017-05-24
Change runserver_plus print syntax for python3 compatibility

### LINKED ISSUE #970: runserver_plus command, pm (postmortem) mode does not work with python 3
Issue is on Line 175 in management/commands/runserver_plus.py:
`print >>> sys.stderr, "Exception occured: %s, %s" % (exc_type, exc_value)`

This syntax does not work with Python 3. To make it work, we need to do:

```
from __future__ import print_function
....
.....
print("Exception occured: %s, %s" % (exc_type, exc_value), file=sys.stderr)

```

Should I send a PR?

cwisecarver:
This fixes #970. It also sorts the imports in that file. LMK if the isort is a problem and I'll remove it and just put the syntax change in.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -1,4 +1,6 @@
 # -*- coding: utf-8 -*-
+from __future__ import print_function
+
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


trbs:
Thanks !

PR #1059, opened 2017-05-26
Add ptpython to the shell_plus docs

keybits:
This is already implemented in https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/shell_plus.py but not documented. This PR corrects that.
I tested the resolution order to confirm this update is correct.
I also changed the capitalisation on bpython since that's how the project refers to itself: https://bpython-interpreter.org/about.html


trbs:
Thanks 1

Issue #1061, opened 2017-05-31
MySQL users will have no alternative when support for JSONField is dropped


RevolutionTech:
I noticed that the current docs say the following with respect to `django-extensions`'s `JSONField`:

> Deprecated since version 1.7.3: Django 1.9 features a native JSONField. Django-Extensions will support JSONField at the very least until Django 1.8 becomes unsupported.

Ref: https://django-extensions.readthedocs.io/en/latest/field_extensions.html#current-database-model-field-extensions

However, the `JSONField` provided by Django is for PostgreSQL only.
Ref: https://docs.djangoproject.com/en/1.11/ref/contrib/postgres/fields/#jsonfield

Thus, when `django-extensions` drops support for `JSONField`, users of MySQL and other database backends will not have an alternative field that they can migrate to. Shouldn't support for this field be reconsidered?

Thank you for your time!

trbs:
I'm happy to leave it in as long as there are (enough) users that find it useful.

Maybe we should improve the documentation instead of deprecating it. Make it very clear in the docs that alternatives exists like the build-in JSONField when using PostGreSQL and maybe some of the other implementations out there which might be better supported.

Of course everybody is happy to use our version :)

RevolutionTech:
That is great news! Thank you very much for considering leaving it in.

trbs:
Would you like to make a PR for this ?

RevolutionTech:
Sure thing! I'll try to tackle that within the next week.

trbs:
Thanks !

PR #1063, opened 2017-06-09
Remove JSONField from deprecation path

RevolutionTech:
Also add note to docs about native JSONField being preferred for users of Django 1.9 and PostgreSQL.

Resolves https://github.com/django-extensions/django-extensions/issues/1061

### DIFF
--- a/django_extensions/db/fields/json.py
+++ b/django_extensions/db/fields/json.py
@@ -14,7 +14,6 @@ from __future__ import absolute_import
 
 import json
 import six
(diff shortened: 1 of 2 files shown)


RevolutionTech:
One of the builds failed:
https://travis-ci.org/django-extensions/django-extensions/builds/241387849

But it looks to be a transient error. I don't have permissions to rebuild this one, unfortunately.

trbs:
Thanks !

PR #1067, opened 2017-06-25
Use 'display_size', not 'precision' to identify MySQL bool field

dsanders11:
Was getting columns not recognized as `bool` on MySQL 5.6.34, looks like the `precision` field is 3 when the code was trying to match up to `tinyint(1)`. Seems like `display_size` might be a more reliable way to detect `tinyint(1)`.


trbs:
Thanks !

PR #1071, opened 2017-07-03
reset_db command: Fix problem that 'utf8_support' option is ignored

mozisan:
Always when I run `reset_db` command, `CHARACTER SET utf8` is append to `CREATE TABLE` statement whether passing `--no-utf8` option or not.
This option is handled at [this code](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/reset_db.py#L125), then fixed this line.


trbs:
Thanks !

PR #1076, opened 2017-07-12
Fix pipchecker "up to date"

vstoykov:
When requirement is from GitHub correctly check commit sha to see if the requirement is up to date if in the requirement string there is branch name instead of commit sha.

The variable `frozen_commit_sha` actually holds the value in the requirement which can be commit hash, branch name or a tag. If it is a branch name or a tag then `frozen_commit_sha != frozen_commit_data["sha"]` where `frozen_commit_data["sha"]` is the actual commit hash of the last commit in the branch or the commit hash marked by a tag.

Currently even if this is the last commit in a branch `pipchecker` will still say that `givenhash is not the head of any branch`. This patch will fix it.


trbs:
Thanks !

PR #1079, opened 2017-07-14
Mention Autocomplete classes for Stacked and Tabular inlines

jackboot7:
Mentions the **ForeignKeyAutocompleteStackedInline** and **ForeignKeyAutocompleteTabularInline** in the documentation. An usage example is still missing.


trbs:
Thanks !

PR #1080, opened 2017-07-19
Update runserver_plus message to account for https

Naddiseo:
This change makes the startup message correctly print "https" vs "http" if  `ssl_context` is available


trbs:
Could you please fix the missing comma so we can merge the PR ?
Thanks.

trbs:
Thanks !

Issue #848, opened 2016-03-03
manage.py print_settings not working with specified names


SeanHayes:
Running Django 1.8.9.

```
$ python manage.py print_settings DEBUG INSTALLED_APPS
usage: manage.py print_settings [-h] [--version] [-v {0,1,2,3}]
                            [--settings SETTINGS]
                            [--pythonpath PYTHONPATH] [--traceback]
                            [--no-color]
manage.py print_settings: error: unrecognized arguments: DEBUG INSTALLED_APPS
```

trbs:
This works for me on both Django 1.8 and Django 1.9.
Are you sure you are running a recent version of django-extensions ?

SeanHayes:
I'm running django-extensions==1.6.1, that's the most recent, right?

SeanHayes:
When I ran the command without args (to print everything) I got a JSON serialization error, if that helps.

Eimis:
I second this, it doesn't work with Django 1.9.8 and django-extensions 1.6.7:

```
(virtualenv)MacBook-Air-2:mijnstroom eimantas$ python manage.py print_settings DEBUG
usage: manage.py print_settings [-h] [--version] [-v {0,1,2,3}]
                                [--settings SETTINGS]
                                [--pythonpath PYTHONPATH] [--traceback]
                                [--no-color] [--format FORMAT]
                                [--indent INDENT]
manage.py print_settings: error: unrecognized arguments: DEBUG
(virtualenv)MacBook-Air-2:mijnstroom eimantas$
```

trbs:
Please test with the latest version on github. This should be solved.

One thing is that right now there are a lot of changes in the github version compared to the last release and I have not had the time to test everything and feel good about making a new release.

If people want to help by testing the github version, trying out the commands (and fields) and report back if they had any issues (or not) it would be much appreciated.

Lrcezimbra:
Any updates on this?

I had the same error with `Python 2.7.13`, `Django==1.10.7` and `django-extensions==1.8.1`

```sh
$ python manage.py print_settings EMAIL_HOST_USER
usage: manage.py print_settings [-h] [--version] [-v {0,1,2,3}]
                                [--settings SETTINGS]
                                [--pythonpath PYTHONPATH] [--traceback]
                                [--no-color] [--format FORMAT]
                                [--indent INDENT]
manage.py print_settings: error: unrecognized arguments: EMAIL_HOST_USER
```

trbs:
[...]

PR #1085, opened 2017-08-09
Fix print_settings to receive positional args

Lrcezimbra:
Issue #848
Fix print_settings to receive positional args and add some tests.


trbs:
Thanks a lot !!
Great that in addition to the fix, we now have tests for this !

I fixed the testing errors which are not related to this PR.

Will do a new release of `django-extensions` shortly so people can update their requirements as needed.

Issue #1089, opened 2017-08-28
graph_models with --json option crashes with TypeError


bijanvakili:
The `--json` option for the `graph_models` command is currently broken and will always output:

```
TypeError: ... is not JSON serializable
```

The command needs to be updated to flatten the context if `--json` is enabled prior to serialization.

trbs:
Thanks !

PR #1090, opened 2017-08-28
Issue 1089: Fix --json option in graph_models

bijanvakili:
Fix `--json` option in graph_models to flatten contexts prior to serialization

See issue #1089


trbs:
Thanks !

PR #1093, opened 2017-09-03
Updating docs for model extensions

fbidu:
Hello, first time contributing here!

I was reading through the code for the Abstract Base Class defined for Django models and noticed that the docs were outdated so I took some time to improve them.

Hope it suits well!


trbs:
Thanks !

Your help is much appreciated.

Issue #1094, opened 2017-09-09
jquery.autocomplete Return: URIError: Malformed URI


smatht:
Failed to get a string that contains the percent symbol (%) in its contents.

Return: URIError: Malformed URI

PR #1095, opened 2017-09-09
Fix issue #1094

smatht:


trbs:
Could you please explain the fix some more ?

It seems like this nullifies `decodeURIComponent` so might as well remove both `URIComponent`'s. But we need to be sure this is the correct fix without side effects.

trbs:
Closing because of inactivity. Please submit a new PR if this is still an issue and somebody wants to pick it up.

PR #1096, opened 2017-09-11
Generate HTML coverage report by default

kevgathuku:
HTML coverage helps a lot during development. It would be nice to have it included by default.


trbs:
Thanks

PR #1098, opened 2017-09-11
Handle runscript import error messages consistently on Python 2 and 3

kevgathuku:
[`importlib.import_module`](https://docs.python.org/3/library/importlib.html#importlib.import_module) in Python 3.3+ works by trying to import all parent packages first before importing a module.
This ended up showing many more errors on Python 3 than on Python 2. 
This makes the error messages shown to the user on an invalid package import consistent on all supported Python versions

### DIFF
--- a/django_extensions/management/commands/runscript.py
+++ b/django_extensions/management/commands/runscript.py
@@ -89,13 +89,22 @@ class Command(EmailNotificationCommand):
                     if not isinstance(e, CommandError):
                         raise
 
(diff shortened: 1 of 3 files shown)


kevgathuku:
Hi @trbs this handles the pending issues from #964 and adds more thorough tests

trbs:
Thanks again ! :-)

kevgathuku:
Happy to help 😄

PR #1100, opened 2017-09-17
Use text/plain for foreign key autocomplete admin

### LINKED ISSUE #1056: Use text/plain for foreign key autocomplete admin
My admin site uses `ForeignKeyAutocompleteAdmin`. One day, our autocomplete fields started looking like this:

screen shot 2017-05-18 at 3 25 59 pm

The full HTML response looked like this:

`<html><head></head><body>My Venue</body></html>`

This issue was that we are using `htmlmin.middleware.HtmlMinifyMiddleware`, and since the response's content type is `text/html`, the response was 'minified' to include HTML tags

I solved this issue by doing this in my admin class:

```
    def foreignkey_autocomplete(self, request):
        response = super(BaseModelAdmin, self).foreignkey_autocomplete(request)
        # HtmlMinifyMiddleware does not play nicely with ForeignKeyAutocompleteAdmin - force response to plain text
        response['Content-Type'] = 'text/plain'
        return response
```

Though this code solved my issue, I suggest that the content type of `foreignkey_autocomplete` response should be set to `text/plain` instead of using the default `text/html` to avoid more issues like this.

jpnauta:
Fixes #1056.

### DIFF
--- a/django_extensions/admin/__init__.py
+++ b/django_extensions/admin/__init__.py
@@ -120,7 +120,7 @@ class ForeignKeyAutocompleteAdminMixin(object):
                     pass
                 else:
                     data = to_string_function(obj)
-            return HttpResponse(data)
+            return HttpResponse(data, content_type='text/plain')
         return HttpResponseNotFound()
 
     def get_related_filter(self, model, request):


trbs:
Thanks !

Issue #1091, opened 2017-08-29
exec in subfunction = syntaxerror


joej:
command "./manage.py shell_plus" throws:
SyntaxError: unqualified exec is not allowed in function 'get_plain' it contains a nested function with free variables

in line 230 of shell_plus.py, this seems to solve it:
                    exec(compile(pythonrc_code, pythonrc, 'exec'), imported_objects) in globals(), locals()

trbs:
Please make a pull request.

Could you please also look into a test for this ? we don't have any tests for `shell_plus` at the moment. It would be very good to hit this with a test so we can make sure there no regressions in the future.

trbs:
I think this is already fixed in git. 
If the mainline version is still not working for you please reopen the ticket and add a test to the test suite.

ewjoachim:
Hi, do you think it would be possible to make a new release with the fix ?

Thank you.

trbs:
Sure will be out shortly 👍

trbs:
The new release does not include any patch for this issue.

If 1.9.1 still causes issues then please add more information about which version of Python and Django Extensions your using and submit a PR with a test for this issue.

Thanks

clokep:
I'm still seeing this with django-extensions 1.9.1 and Python 2.7.6 with Django 1.10.8.

Full stack trace:
```
Traceback (most recent call last):
  File "manage.py", line 45, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/usr/local/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 208, in fetch_command
    klass = load_command_class(app_name, subcommand)
  File "/usr/local/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 40, in load_command_class
    module = import_module('%s.management.commands.%s' % (app_name, name))
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/usr/local/venv/local/lib/python2.7/site-packages/django_extensions/management/commands/shell_plus.py", line 229
    exec(compile(handle.read(), pythonrc, 'exec'), imported_objects)
[...]

PR #1102, opened 2017-09-21
Fixes shell_plus command SyntaxError on exec(), python compatibility

ppython:
Fixes the SyntaxError in the shell_plus command.
It seems users with python 2.x has this problem.

I tried implemented @joej simple solution from https://github.com/django-extensions/django-extensions/issues/1091#issue-253504617

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -226,7 +226,7 @@ class Command(BaseCommand):
                 # PYTHONSTARTUP prints an exception and continues.
                 try:
                     with open(pythonrc) as handle:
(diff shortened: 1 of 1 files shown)


trbs:
Could you please try to add tests for this ?

tuky:
It's a bug in python, which was fixed in 2.7.9. See https://github.com/pyinstaller/pyinstaller/issues/1408#issuecomment-132014629 and https://bugs.python.org/issue21591. So it would be easy to create a test, which does nothing, but just import this management command module. But I don't know, if that would help a lot. It would be just a syntax check. So I would love to see this merged, but you should also upgrade python.

trbs:
Thanks that explains a lot.

PR #1112, opened 2017-10-20
Skipping all models importing feature added.

domandinho:
Phase of importing models can be skipped through command line:
`python manage.py shell_plus --dont-load=*`
or in django settings:
`SHELL_PLUS_DONT_LOAD=['*']`

Logic of importing models is unchanged. It is now under if statement.


trbs:
Thanks !

PR #1113, opened 2017-10-22
Script execution directory policy feature added.

domandinho:


domandinho:
CI will be fixed soon.

trbs:
Thank you so much !

trbs:
I was thinking:

Should we also add a `settings.RUNSCRIPT_DIR_POLICY` ? so that users do not have to set `--dir-policy` every time they call `runscript`.

And maybe `--chdir` should just always override the working directory ?

Issue #1104, opened 2017-10-02
jupyter lab


strelnikov:
please consider jupyter lab for shell_plus ;)

trbs:
I would love to be able to support JupyterLab !
(and in general improve our Jupyter notebook support)

PR's are most welcome !

bhrutledge:
@strelnikov Here's a temporary workaround:

Assuming you've run `manage.py shell_plus --notebook` at least once, you can [install JupyterLab](https://github.com/jupyterlab/jupyterlab#installation) in your virtual environment, and then just run `jupyter lab`. You should be able to create new "Django Shell-Plus" notebooks, and run existing ones. This works because `shell_plus --notebook` installs a Jupyter kernel, which is left in place even after you kill `shell_plus`.

bhrutledge:
As I understand it, this is a request for something like `manage.py shell_plus --lab`.

Another way to approach this would be to encourage folks to install Jupyter notebook/lab/etc. into a separate environment, and then have a new management command for installing a kernel for the current Django environment. This seems to follow the recommendation from the [IPython docs](https://ipython.readthedocs.io/en/latest/install/kernel_install.html#kernels-for-different-environments). I think the behavior would be similar to `python -m ipykernel install --user --name myvenv`, but with a kernel spec that has an `argv` similar to the current `shell_plus --notebook`.

FWIW, I have a repo for my base Jupyter environment, with instructions on how to add kernels for other environments: https://github.com/bhrutledge/jupyter-venv. The advantage of this approach is that it avoids cluttering up the Django environment with all of Jupyter's dependencies, and means you only have to have one notebook/lab instance running. Of course, the current approach of `shell_plus --notebook` is probably easier for people who haven't used Jupyter before.

jdmold:
@n1-cz if you want something quick and dirty install jupyter lab, run `manage.py shell_plus --notebook`. Then open the notebook url, get the sign in out of the way, and go to `/lab`. You'll see the django shell in launcher, I've used it a bit everything seems to work.

Lets say I'm running `docker-compose run django-container -p 36000:8888 manage.py shell_plus --notebook` after taking care of the token stuff I then go to `localhost:36000/lab` and bask in the goodness.

![screen shot 2018-10-26 at 3 35 55 pm](https://user-images.githubusercontent.com/7205787/47591343-df23eb00-d934-11e8-99c7-8d31c10cc71f.png)

kgang:
[...]

PR #1120, opened 2017-11-03
Install notebook kernel spec to virtual environment

### LINKED ISSUE #1026: Conflicting Jupyter kernel specs from multiple environments
I'm developing multiple Django projects, each with its own virtualenv that includes django-extensions. To user the Jupyter Notebook, I've been doing `pip install jupyter` followed by `manage.py shell_plus --notebook` inside the virtualenv. It's been working okay, but I've learned a little bit more about how Jupyter works, and it seems that the way django-extensions handles kernel specs is not ideal.

For example, when I run the notebook in `project-foo`, I get the following kernel spec at `/Users/brian/Library/Jupyter/kernels/django_extensions/kernel.json`:

```
{"language": "python", "display_name": "Django Shell-Plus", "env": {"PYTHONPATH": "/Users/brian/Code/project-foo"}, "argv": ["/Users/brian/.virtualenvs/project-foo/bin/python", "-m", "ipykernel", "-f", "{connection_file}", "--ext", "django_extensions.management.notebook_extension"]}
```

However, when I run the notebook in `project-bar`, that overwrites the *same file* with:

```
{"language": "python", "display_name": "Django Shell-Plus", "env": {"PYTHONPATH": "/Users/brian/Code/project-bar"}, "argv": ["/Users/brian/.virtualenvs/project-bar/bin/python", "-m", "ipykernel", "-f", "{connection_file}", "--ext", "django_extensions.management.notebook_extension"]}
```

Now, in practice, this hasn't been a huge problem, because I'm usually not running multiple notebook servers at once, and django-extensions always writes the kernel spec before it starts a notebook server. But it does seem like it would cause unexpected and undesirable behavior.

After reading up on [installing kernels](https://ipython.readthedocs.io/en/latest/install/kernel_install.html), I see two potential fixes:

* Instead of installing to the [user's kernel directory](https://github.com/django-extensions/django-extensions/blob/1.7.6/django_extensions/management/commands/shell_plus.py#L195), install to `sys.prefix`. IE, mimic `python -m ipykernel install --sys-prefix` instead of `--user`. So, in the example above, the kernel spec for `project-foo` would be installed to `/Users/brian/.virtualenvs/project-foo/share/jupyter/kernels/django-extensions`.
* Mimic the behavior of the `--name` option to `python -m ipykernel install`, similar to the [option for setting `display_name`](https://github.com/django-extensions/django-extensions/blob/1.7.6/django_extensions/management/commands/shell_plus.py#L224-L225). So, in the example above, the kernel spec for `project-foo` would be installed to `/Users/brian/Library/Jupyter/kernels/django_extensions-project-foo` (possibly with a `display_name` of `Django Shell-Plus (project-foo)`.

bhrutledge:
Quick fix #1026.

I don't love using the private [`_get_destination_dir`](https://github.com/jupyter/jupyter_client/blob/2822872afb8fc52b44cf9e18febe68d1375124a2/jupyter_client/kernelspec.py#L247), but it gets the job done. I'd like to go further and use the [public API for `KernelSpecManager`](https://jupyter-client.readthedocs.io/en/stable/api/kernelspec.html#jupyter_client.kernelspec.KernelSpecManager), which installs kernel specs from a directory. I'd probably do something similar to the [implementation of `python -m ipykernel install`](https://github.com/ipython/ipykernel/blob/9e7d1920e376de4e290009a760464b948a09d85c/ipykernel/kernelspec.py#L86-L88), which writes the spec to a temporary directory, installs it, then removes the directory. But, before I do that, I thought maybe a PR discussion might be in order.

Also, for people who have already run `manage.py shell_plus --notebook`, the kernel spec installed in `user_kernel_dir` will take precedence over this. Ideally, it should be deleted, but I'm wary of doing that automatically. What if I added a warning along the lines of `The kernel installation location has changed; please delete {ksm.user_kernel_dir}/django_extensions`?

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -160,7 +160,8 @@ class Command(BaseCommand):
 
                 ks.env['PYTHONPATH'] = ':'.join(filter(None, pythonpath))
 
-            kernel_dir = os.path.join(ksm.user_kernel_dir, 'django_extensions')
+            # Install the kernel to the virtual environment
+            kernel_dir = ksm._get_destination_dir('django_extensions', prefix=sys.prefix)
             if not os.path.exists(kernel_dir):
                 os.makedirs(kernel_dir)
             with open(os.path.join(kernel_dir, 'kernel.json'), 'w') as f:


trbs:
Doesn't this potentially fail for people not using virtualenv ?
`sys.prefix` on system installed Python is likely not a writable directory for normal users.

Would it make sense to make a command line argument and/or settings variable for this ?
So users can chose between installing in virtualenv and `user=True` (`user_kernel_dir`).

--

I think using the public API for `KernelSpecManager` would be awesome. 
Would that also help with https://github.com/django-extensions/django-extensions/issues/1104 ?

bhrutledge:
I think this will work for anyone who was able to do `pip install django_extensions`, which would install to `sys.prefix`. Also, as described in the original issue, [always installing the kernel as `django_extensions`](https://github.com/django-extensions/django-extensions/blob/9cea102ccfe376fdd5e6d1b933b46e4c1c4085a9/django_extensions/management/commands/shell_plus.py#L163) means you can really only have one kernel at a time per Python environment.

I think properly installing it to `user_kernel_dir` would require a unique kernel name, but that plus adding a setting or command line argument seems like extra work for something that will be confusing to people who aren't familiar with kernel specs. The value in installing to `user_kernel_dir` is that it's accessible to `jupyter` running in a different environment, but `shell_plus --notebook` runs the notebook in the same environment as `django`, so I think it makes sense to install the kernel in the same place.

All that said, I think it'd at least be worth adding a helpful error message if the kernel installation fails, and maybe even some additional info messages like `Installing Django Shell-Plus kernel to {sys.prefix}...`. But I'll also dig further into the kernel spec API for more ways to improve this. I don't think this really addresses #1104, which I understand to be a request for "shell_plus --lab".

bhrutledge:
Another idea, borrowed from the [environment_kernels](https://github.com/Cadair/jupyter_environment_kernels) package:

Instead of writing a kernel spec to disk, we could subclass [`KernelSpecManager`](https://github.com/jupyter/jupyter_client/blob/2822872afb8fc52b44cf9e18febe68d1375124a2/jupyter_client/kernelspec.py#L108) to return a `django_extensions` spec for the current environment as a `dict` or `KernelSpec` where appropriate. I still have to figure out how to wire that up to the `NotebookApp`, but it seems doable, and avoids the issues around permissions and multiple environments.

trbs:
Sounds good ! 👍 

I'm more then happy to change the way we handle this even if it's a backwards incompatible / breaking change if that means we can improve notebook handling :)

bhrutledge:
Well, if backwards-incompatible/breaking changes are on the table... 😉 

What'd I'd really like to do is something like what I [suggested on #1104](https://github.com/django-extensions/django-extensions/issues/1104#issuecomment-341847301). More specifically, I'd remove the `--notebook` and `--kernel` options for `shell_plus`, and add a new command, maybe just called `install_kernel`. This would behave and have options similar to [`python -m ipykernel install`](https://ipython.readthedocs.io/en/latest/install/kernel_install.html#kernels-for-different-environments)  and the [bash kernel](https://github.com/takluyver/bash_kernel) with respect to installing a kernel spec, but would add the `django_extensions` notebook extension. I'd try to have helpful defaults, like defaulting to `--user` with a name that reflects the active virtual environment.

With the kernel spec installed, users would be free to run `jupyter notebook`, `jupyter lab`, `jupyter qtconsole`, etc. in whatever environment they want (their Django virtualenv, a Jupyter virtualenv, Anaconda, etc.). And, `django_extensions` wouldn't need the `NOTEBOOK_ARGUMENTS` and `IPYTHON_ARGUMENTS` settings, because it's no longer acting as a wrapper around the `notebook` command, and users would be free to edit the installed kernel spec as needed.

It's a big change, but it seems more in line with how Jupyter kernels are designed to work. I'm happy to take ownership of this, including updating the docs. What do y'all think?

trbs:
Think that sounds like a plan ! also with a new command we can have a clear upgrade path outlined in the docs. (Something like; introducing the new command, recommending a switch, then later deprecating `--notebook` and eventually removing it)

bhrutledge:
Awesome. I'll get to work. Closing this PR.

PR #1121, opened 2017-11-04
validate_templates improvement

jonashaag:
Allow --include paths like "app/templates".

Otherwise there's no way to include app templates (only template folders in your project root).


trbs:
Thanks you for the PR request.

I hope the following two commits solve your problem in a slightly different way:

https://github.com/django-extensions/django-extensions/commit/ea2f71a9045dac7b36d9e193bcb844795d67a781
https://github.com/django-extensions/django-extensions/commit/bba0ef9f4c3d89b507f65193555600e3c4a5d76d

If not feel free to reopen.

Issue #892, opened 2016-05-11
Wrong search_path for autocomplete widget


lsgd:
I'm using Django 1.9.6 and django-extensions 1.6.7

In my code, the widget.py generated a wrong URL for the ajax request which lead to a redirect to modify an object with ID "foreign_key".

After some investigation I figured out, that the generated URL uses the model_name of the foreign key model instead of the form model:

```
# models.py
class Order(models.Model):
    customer = models.ForeignKey(Customer)
    ...
    class Meta:
        app_label = 'crm'

# admin.py
class OrderAdmin(ForeignKeyAutocompleteAdmin):
    list_display_links = None
    list_display = ('customer_name', ... )
    list_per_page = 25
    inlines = [ OrderPositionInline]
    form = CustomOrderForm

    related_search_fields = {
        'customer': ('firstname', 'lastname'),
    }
```

The generated URL in **django_extensions/admin/__init__.py** is **_/admin/crm/order/foreignkey_autocomplete/**_

The generated URL in **django_extensions/admin/widgets.py** is **_/admin/crm/customer/foreignkey_autocomplete**_

In **widgets.py** the model_name is extracted from the **to** part of the relation in [line 45](https://github.com/django-extensions/django-extensions/blob/9b5a18eb8329e3a38d583ba82f7a2c90f6945423/django_extensions/admin/widgets.py#L45) and used in the base url **related_url** in [line 48](https://github.com/django-extensions/django-extensions/blob/9b5a18eb8329e3a38d583ba82f7a2c90f6945423/django_extensions/admin/widgets.py#L48).

```
def render(self, name, value, attrs=None):
    if attrs is None:
        attrs = {}
    opts = self.rel.to._meta

    app_label = opts.app_label
    model_name = opts.object_name.lower()

    related_url = reverse('admin:%s_%s_changelist' % (app_label_url, model_name_url))
```

I fixed the problem by retrieving the model_name and app_label from the model on which the field is located in. The model_name of the related **to** field is used later as GET parameter.

```
opts = self.rel.to._meta
app_label = opts.app_label
model_name = opts.object_name.lower()

opts_url = self.rel.field.model._meta
app_label_url = opts_url.app_label
model_name_url = opts_url.object_name.lower()

related_url = reverse('admin:%s_%s_changelist' % (app_label_url, model_name_url))
```

Do I really need this modification or did I just miss a flag or parameter to manually set the correct model?

If this is a bug, shall I create a pull request?

[...]

PR #1126, opened 2017-11-17
Apply fix described in #892

reavis:
Fix the URL used in the ForeignKeySearchInput widget.


trbs:
Could you please write a test that (at least) follows the code path in `def render(self, name, value, attrs=None):` ? 

That way we have the PR tested on all supported Django versions and will catch any potential issues in the future :-)

Thanks !

PR #1128, opened 2017-11-20
fixes #1127 -- pipchecker can't find some packages

### LINKED ISSUE #1127: pipchecker doesn't find misconfigured pypi packages
I've encountered problems with running pipchecker when there's a package which is misconfigured on pypi with underscores instead of dashes.  For example:

```
% manage.py pipchecker
django-celery-beat 1.0.1         release is not on pypi (check capitalization and/or --extra-index-url)
django-polymorphic 1.3           release is not on pypi (check capitalization and/or --extra-index-url)
```

Each of these packages seems to have been set up incorrectly to use underscores instead of dashes, and I've filed bugs for both of these packages to change their pypi entries (hopefully)
https://github.com/django-polymorphic/django-polymorphic/issues/319#issuecomment-345702344
https://github.com/celery/django-celery-beat/issues/85

But I also think it would be great if pipchecker could be more resilient.  I've tested a fix that seems to work which when searching the parsed pypi xml data searches for both the proper name, and a transformed name where dashes are replaced by underscores.

`available = pypi.package_releases(req["pip_req"].name, True)`
->
`available = pypi.package_releases(req["pip_req"].name, True) or  pypi.package_releases(req["pip_req"].name.replace('-','_'), True)`

picturedots:
https://github.com/django-extensions/django-extensions/issues/1127

### DIFF
--- a/django_extensions/management/commands/pipchecker.py
+++ b/django_extensions/management/commands/pipchecker.py
@@ -136,7 +136,7 @@ class Command(BaseCommand):
             elif "dist" in req:
                 dist = req["dist"]
                 dist_version = LooseVersion(dist.version)
-                available = pypi.package_releases(req["pip_req"].name, True)
+                available = pypi.package_releases(req["pip_req"].name, True) or pypi.package_releases(req["pip_req"].name.replace('-','_'), True)
                 available_version = self._available_version(dist_version, available)
 
                 if not available_version:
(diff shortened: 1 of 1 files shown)


trbs:
Could you please fix the style error ?

What are the odds that by replacing `-` with `_` we match a different package ? (or does `pypi` protect against that ?)

trbs:
Thanks !

Guess `pipchecker` in itself will not create a security thread. Thinking if a user might be tempted to install/upgrade to a typo-squatted named version if that's detected as a newer version.

PR #1134, opened 2017-12-04
Update test matrix

michael-k:
Django 2.0 has been released and 1.10 reached its EOL: https://www.djangoproject.com/weblog/2017/dec/02/django-20-released/


trbs:
Thanks !

PR #1137, opened 2017-12-08
Configuring paths to certificates in runserver_plus added.

domandinho:


docs/runserver_plus.rst
```diff
@@ -100,7 +100,7 @@ runserver_plus also supports SSL, so that you can easily debug bugs that pop up
 when https is used. To use SSL simply provide a file name for certificates;
 a key and certificate file will be automatically generated::
 
-  $ python manage.py runserver_plus --cert cert
+  $ python manage.py runserver_plus --cert-file cert
```
trbs: Think we should use `--cert-file cert.crt` here.

I would prefer we use the full paths since they are more clear and easier to understand. We can explain under `Certificate Paths` that a path to the base name is also accepted as a shorthand.

django_extensions/management/commands/runserver_plus.py
```diff
@@ -401,7 +434,6 @@ def check_migrations(self):
                 self.stdout.write(self.style.NOTICE("\nYou have unapplied migrations; your app may not work properly until they are applied."))
                 self.stdout.write(self.style.NOTICE("Run 'python manage.py migrate' to apply them.\n"))
```
trbs: Small style error here with missing extra newline.

trbs:
Thanks for the PR ! Few small tweaks above, lets get this merged 👍

trbs:
Awesome many thanks !

PR #1138, opened 2017-12-13
Create a management command for de-duplicating model instances.

YPCrumble:
First, thanks for building this great library. I use it in every project I build.

This PR creates a management command for merging duplicate model instances on a specified field(s). I frequently find myself having to merge (or de-duplicate) model instances on existing Django sites.

There are various gists out there, including the three I used for inspiration (see the comments on the command), but all of these were buggy and kept getting outdated.

The command asks the user to input the model to deduplicate, the field on which to de-duplicate the model, and whether the user wants to keep the first (or last) duplicate instance of the model. The command then merges all duplicate instances of the model into the first (or last) instance. Related fields (M2M, FK) of all duplicate model instances are also merged onto the remaining instance.

Thanks for taking a look - I hope this will be considered a helpful addition to django-extensions! Please let me know if I can add any improvements or revisions.

### DIFF
--- a/django_extensions/management/commands/merge_model_instances.py
+++ b/django_extensions/management/commands/merge_model_instances.py
@@ -1,0 +1,225 @@
+# -*- coding: utf-8 -*-
+from django.apps import apps
+from django.contrib.contenttypes.fields import GenericForeignKey
+from django.core.management import BaseCommand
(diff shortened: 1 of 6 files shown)


trbs:
Thanks !

PR #1142, opened 2017-12-20
Add pluggable dev command

jpic:
I use this command on a daily basis, it can run a development server after automatically migrating the database and creating some users with a random password, also, it has a pluggable threading server to plug your other commands (test watch, linter watch, static watch ...)

### DIFF
--- a/django_extensions/management/commands/dev.py
+++ b/django_extensions/management/commands/dev.py
@@ -1,0 +1,202 @@
+"""
+Pluggable development environment.
+
+Example to add a thread to this command, somewhere that will be imported at
+startup (ie. yourproject/models.py if you added yourproject to
+INSTALLED_APPS)::
+
+    import os
+
+    from crudlfap.management.commands.dev import Command, CommandThread
+
+    Command.threads.append(
+        CommandThread(
+            name='npm-watch',
+            cmd='npm start -- --watch',
+            cwd=os.path.abspath(
+                # should be path to package.json
+                os.path.join(os.path.dirname(__file__), '..', '..')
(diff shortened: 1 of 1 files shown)


jpic:
Quick question: is it ok to support anything starting from py36 and dj20 to contribute a command to django-extensions ? Thanks <3

trbs:
Very interesting addition, many thanks for the PR !

It would be nice to make it fully configurable for users. For example the dev/staff/user accounts are hardcoded. 

Also think everything should be configurable from `settings.py` (with sane defaults), from the top of my head:

- the users in `def createusers(self)` from a list/dict in `settings.py`
- the threads to run from a list/dict in `settings.py`
- instead of hardcoded `self.createusers()` in `def handle(self, *args, **options):` use a for-loop just like `self.threads` configured from a list/dict in `settings.py`
- for the above move `def createusers(self):` to a function instead of a method

As for supporting only Python 3.x and Django 2.x, I strongly prefer we only add commands which are fully supported following our support schedule. Think Django-Extensions should be conservative with dropping support for older Python and Django versions. Specially since we do not have the funds or men power to do separate bug fix and feature releases for multiple sets of Python / Django releases.

Lastly please add tests and documentation for the dev command.

Sorry for the long list of questions and improvements, I do think it's a very cool command and can make life so much easier for all kind of Django developers.

Would love to include it !

jpic:
It's true this command is a lot of fun to use, but we should be able to print out a new password for any user with a keypress and switch between windows, perhaps with https://github.com/jonathanslenders/pymux ?

PR #1143, opened 2017-12-21
Use tuples for `startswith` and `endswith` calls instead of multiple calls.

Stranger6667:
This approach is more readable and a bit faster


trbs:
thanks !

PR #1144, opened 2017-12-22
Fixes #928 AutoSlugField.deconstruct does not match construction values

### LINKED ISSUE #928: AutoSlugField.deconstruct does not match construction values
I'll try to submit a PR later but in case i forget:

The deconstruct() call should return the same which were used to initalize the field. However the populate_from property is changed from create_slug() which causes the django migration checker to think that a migration is required.

See workaround solution here: https://github.com/django-oscar/django-oscar/commit/27cb08936252ae320423ecccc22b5658135c5b04

tomviner:
We also hit the problem described in https://github.com/django-oscar/django-oscar/commit/27cb08936252ae320423ecccc22b5658135c5b04 where a unittest to detect missing migrations keeps failing because if the test run creates a slug, the `_populate_from` attr gets changed, and this alters the result of `deconstruct`, so Django wants another migration to include this change.

This change:
- ensures the values that come out of `deconstruct()` are unchanging
- the value of `populate_from` matches that values passed into `__init__`

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -149,14 +149,15 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
 
     def create_slug(self, model_instance, add):
         # get fields to populate from and slug field to set
-        if not isinstance(self._populate_from, (list, tuple)):
-            self._populate_from = (self._populate_from, )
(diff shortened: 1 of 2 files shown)


trbs:
Thanks !

PR #1148, opened 2018-01-17
Fix runserver_plus in Django 1.11 using MIDDLEWARE_CLASSES

bhrutledge:
We're upgrading to Django 1.11, but not (yet) using `settings.MIDDLEWARE`. However, at runtime it defaults to `None`, which results in `TypeError: argument of type 'NoneType' is not iterable` when
running `runserver_plus`. This approach allow for both styles of middleware in Django < 2.0.

I've only tested this manually in our 1.11 environment.


bhrutledge:
Modification of fix for Django 2.0 in b6951d39736fc1d13e667c3a11f52f5c1764767d.

trbs:
Thanks !

PR #1149, opened 2018-01-19
Fix display of show_urls command in Django 2.0

sciyoshi:
Currently, output of `show_urls` looks like this (notice the extra single quotes):

screenshot from 2018-01-18 22-16-18

`describe()` is not suitable for showing the actual URL - instead, simply convert to a string.


trbs:
Thanks !

PR #1155, opened 2018-02-12
Bug in including proxy models for sqldiff removed

domandinho:


trbs:
Thanks.

Would you be able to also write a test for this ?

domandinho:
@trbs 
There are no tests for sqldiff in this project or i missed them. How they should look?

trbs:
Your right... regretfully there are no tests yet :-(

Also thinking of it, it's probably not trivial to just add a test for this case.

Adding a proxy model to one of the test models is the easy part but I guess for `sqldiff` the easiest way to approach tests would be to setup/migrate a temporary `sqlite` database so there is a database to run tests against.

Luckily for this PR the test does not need to modify the database itself but the setup code for `sqldiff` tests would need to be written to support tests.

PR #1161, opened 2018-02-19
Add collision resolvers based on app label for shell_plus

browniebroke:
To avoid having very long model names, add 2 new collision resolvers `AppLabelPrefixCR` and `AppLabelSuffixCR` leveraging the app label rather than the app name.


trbs:
Thanks !

Issue #1164, opened 2018-02-20
Broken links to Werkzeug documentation


truthdoug:
The documentation for RunServerPlus has two links to Werkzeug documentation that don't work. The URLs have specific versions in the URL path. Without the version ids, the Werkzeug site redirects the user to the current version of documentation.

Patch forthcoming.

trbs:
Thanks for the PR!

PR #1165, opened 2018-02-20
Fixed links to Werkzeug documentation (issue #1164).

truthdoug:
Also cleaned up preface in debugger PIN section.


trbs:
Thanks !!

PR #1171, opened 2018-02-27
Add function support to AutoSlugField populate_from

lskillen:
Hello!

As per the title, adds the ability to specify functions directly in the `populate_from` argument of `AutoSlugField`. For example:

```
def get_readable_title(instance):
    return "The title is {}".format(instance.title)

class FunctionSluggedTestModel(models.Model):
    title = models.CharField(max_length=42)
    slug = AutoSlugField(populate_from=get_readable_title)

    class Meta:
        app_label = 'django_extensions'
```

The motivation for this is that Django disallows methods to be called via migrations (easily), and this proves to be a major inconvenience when the slug generation isn't simply derived. :-)

This PR also pushes the evaluation of whether the slug should be overwritten or not further up, so that when the value is already specified, none of the logic for slug generation is called. This means that you can create objects with slugs pre-populated (e.g. data migration) and you can be certain that none of the logic will be executed. It's also slightly more efficient.

All tests executed successfully apart from unrelated issues with sqldiff tests (on my machine).

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -148,23 +148,22 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
         raise RuntimeError('max slug attempts for %s exceeded (%s)' % (original_slug, self.max_unique_query_attempts))
 
     def create_slug(self, model_instance, add):
+        slug = getattr(model_instance, self.attname)
+        if slug and not self.overwrite:
+            # Existing slug and not configured to overwrite - Short-circuit
+            # here to prevent slug generation when not required.
+            return slug
+
         # get fields to populate from and slug field to set
         populate_from = self._populate_from
         if not isinstance(populate_from, (list, tuple)):
             populate_from = (populate_from, )
         slug_field = model_instance._meta.get_field(self.attname)
 
-        if add or self.overwrite:
-            # slugify the original field content and set next step to 2
-            slug_for_field = lambda lookup_value: self.slugify_func(self.get_slug_fields(model_instance, lookup_value))
-            slug = self.separator.join(map(slug_for_field, populate_from))
-            start = 2
-        else:
-            # get slug from the current model instance
-            slug = getattr(model_instance, self.attname)
-            # model_instance is being modified, and overwrite is False,
-            # so instead of doing anything, just return the current slug
-            return slug
+        # slugify the original field content and set next step to 2
+        slug_for_field = lambda lookup_value: self.slugify_func(self.get_slug_fields(model_instance, lookup_value))
+        slug = self.separator.join(map(slug_for_field, populate_from))
+        start = 2
 
         # strip slug depending on max_length attribute of the slug field
         # and clean-up
@@ -182,6 +181,10 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
             model_instance, slug_field, self.slug_generator(original_slug, start))
 
     def get_slug_fields(self, model_instance, lookup_value):
+        if callable(lookup_value):
+            # A function has been provided
+            return "%s" % lookup_value(model_instance)
+
         lookup_value_path = lookup_value.split(LOOKUP_SEP)
         attr = model_instance
         for elem in lookup_value_path:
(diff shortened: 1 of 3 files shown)


trbs:
Thanks for the PR !

Sorry for not responding sooner. 

The PR looks good ! Are you sure that changing of the order does not have other side effects ?

What I would like to suggest is to split this into two PR's... one for allowing the use of a callable and the other for improving `create_slug`.

Would you be able to do that ?

Thanks in advance !

lskillen:
@trbs We've been successfully using this in production since I created the PR, so it looks good so far. :-)

I don't have an issue with splitting it into 2, but it might have to wait for 4 weeks as I'm completely flat-out next week and then I'm away to China for 3 weeks after that. If I can get the chance next week I'll try to do that, but otherwise it'll be awhile. :-)

trbs:
That's great to hear.

If you don't mind I can split it up in two commits as well to have it merged sooner.

trbs:
As I don't have time for it as well the next few weeks but wanted to get it included I just merged the PR as is 👍 

Thanks for the work on this !

lskillen:
@trbs Thanks! I'm still beyond busy too. If it causes problems, which I really think it won't, I'll personally take the blame. ;-)

tisdall:
@lskillen - prepare to take blame!  :wink: 

This introduced a bug when making a copy of an existing object (by setting `pk` to `None`).  Prior to this patch the `create_slug` would be called with `add` being true and then a new slug would be properly created.  This patch short-circuits and ignores `add` and then the copy is saved with the prior slug value (causing a collision).  In fact, it looks like the new method completely ignores the `add` value.

So, this used to work but now doesn't:

```
x = MyModel(name='hi')
x.save()
x.pk = None
x.save()
```

lskillen:
@tisdall Blame accepted! ;-) Please accept my apologies.

Looks like the tests need some additional test coverage for `add` test case, to include your exact example. On the upside, it looks like it'd be easy to fix - As you've said, it needs to take the value of add into account.

tisdall:
It looks like it's just changing `if slug and not self.overwrite:` to `if slug and not self.overwrite and not add:` and then additional tests.

trbs:
Is there a new pull request for this fix ?

lskillen:
I raised #1217 - Apologies for the wait (especially to @tisdall who was directly impacted). I had intended to fix this almost immediately but other bigger fires were burning at the time (*hangs headi in shame*). Also encountered issues with running the test suite based on master. Lots of new (but unrelated) failures for:

> RuntimeError: Conflicting 'teslacar' models in application 'testapp_with_no_models_file': <class 'tests.testapp_with_no_models_file.admin.TeslaCar'> and <class 'tests.testapp_with_no_models_file.teslacar.TeslaCar'>.

it didn't impact running the tests for this fix though.

trbs:
Thanks @lskillen !

@tisdall could you confirm this works for you as well ? then I will make a new release as soon as possible.

tisdall:
@trbs - Yes, it looks good to me and the test covers the issue I was having.

tisdall:
@lskillen - Thanks for fixing this.  I was already using a wrapper around the class so I was able to work around the issue by overriding the `create_slug` method.

lskillen:
@tisdall - You're very welcome, and apologies it took so long. Fires have been dealt with. :-)

trbs:
https://github.com/django-extensions/django-extensions/issues/1251

Describes an issue following the fix for the bug here when copying a model.

I'm wondering if there is a proper fix for this... something that works for both:
```
Model.objects.create(slug='key', title='title')
```
and
```
x = MyModel(name='hi')
x.save()
x.pk = None
x.save()
```

I am thinking that in the later case it makes sense that one should also explicitly clear the slug, `x.slug = None` as well ?

That would make the intention explicit... otherwise how do we know if this is a new model with the slug explicitly set or a copied model with the same slug (eg if you do versioning or multi-language) or a copied model that should have a new unique slug.

For now I'm leaning towards making the clearing of the slug explicit if you want the copied model to recalculate the slug.

PR #1175, opened 2018-03-07
Update shell_plus.py

Azd325:
Fix TypeError: super() takes at least 1 argument (0 given)

PR #1177, opened 2018-03-08
Fix shell_plus on python2.7

voidus:
This should be compatible with all current python versions


dsanders11:
Looks like this is a dupe of #1175

trbs:
Thanks for the PR ! and sorry for the inconvenience. 

I merged #1175 only because it had a lower PR number ;-)

voidus:
It's all good, just happy to see it fixed :)

PR #1186, opened 2018-03-20
Fix print-sql: call execute on CursorWrapper instead of directly

vesterbaek:
`runserver_plus --print-sql` and `shell_plus --print-sql` have the same implementation for printing sql. They call `execute` on the cursor directly instead of calling `execute` on `django.db.backends.utils.CursorWrapper` (as `django.db.backends.utils.CursorDebugWrapper` does). 

This is a problem because then we do not get the query [executed with wrappers](https://github.com/django/django/blob/51a00749e9d1814acfb6bf8732ecd21f18944083/django/db/backends/utils.py#L68) . => [`DatabaseErrorWrapper` ](https://github.com/django/django/blob/51a00749e9d1814acfb6bf8732ecd21f18944083/django/db/utils.py#L51)is not properly run on db exceptions. A notable effect of the is that Integrity errors are eg. (for Postgres db backend) not converted from `psycopg2.IntegrityError` to `django.db.utils.IntegrityError` which might have a severe impact on code relying on catching `django.db.utils.IntegrityError`.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -157,7 +157,7 @@ class Command(BaseCommand):
                 def execute(self, sql, params=()):
                     starttime = time.time()
                     try:
-                        return self.cursor.execute(sql, params)
+                        return utils.CursorWrapper.execute(self, sql, params)
                     finally:
                         execution_time = time.time() - starttime
                         raw_sql = self.db.ops.last_executed_query(self.cursor, sql, params)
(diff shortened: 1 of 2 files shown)


trbs:
Thanks !

Any reason why we won't want to use `super` instead of `utils.CursorWrapper.execute` ?

vesterbaek:
We do not want to run the implementation in `CursorDebugWrapper. execute` (super) since this prints out debug information. I presume that's why the current implementation directly calls execute on the cursor.

trbs:
Sorry your right I misread `CursorDebugWrapper` vs `CursorWrapper`.

Thanks for the PR !

PR #1189, opened 2018-03-31
FIX bug in windows when PYTHONPATH is defined

gabomgp:
FIX bug in windows when PYTHONPATH is defined and not empty. When working with a django project that use a lib located  in a location outside of the django folder, i need to configure the enviroment variable PYTHONPATH to the path where the library is located. But, in windows, django extensions shell plus is using ":" as path separator and the windows separator is ";". That cause the library can't be located and django is not initializated.


trbs:
Thanks !

PR #1191, opened 2018-04-04
Use introspection module

### LINKED ISSUE #1010: The sqldiff command doesn't recognize GeoDjango and JSON fields
The sqldiff is a great tool.  However, I'm noticing that it does not recognize less common fields, such as PointFields from GeoDjango and JSON fields from the django-jsonfield project. Even if I run the alter statements that sqldiff suggests, sqldiff still reports difference in the output. 

### The sqldiff command doesn't seem to recognize geography columns from GeoDjango:  

Code Sample:

    from django.contrib.gis.db import models as geomodels
    class Location(models.Model):
        ...
        point = geomodels.PointField(editable=False, geography=True, null=True)

Sqldiff output:

    -- Application: eventcalendar
    -- Model: Location
    -- Comment: Unknown database type for field 'point' (130253)
    ALTER TABLE "eventcalendar_location"
    	ALTER "point" TYPE geography(POINT,4326);

### The sqldiff command doesn't seem to recognize geometry columns from GeoDjango:  

Code Sample:

    from django.contrib.gis.db import models as geomodels
    class Office(models.Model):
        ...
        point = geomodels.PointField(editable=False)

Sqldiff output:

    -- Application: physicians
    -- Model: Office
    -- Comment: Unknown database type for field 'point' (129699)
    ALTER TABLE "physicians_office"
	ALTER "point" TYPE geometry(POINT,4326);
    COMMIT;

### The sqldiff command doesn't seem to recognize jsonfields (from django-jsonfield). I haven't tried with the newer built-in jsonfields yet, so maybe it will work with that? 

Code Sample:
https://github.com/ImaginaryLandscape/django_site_config/blob/master/site_config/backends/model_backend/models.py#L110

Sqldiff output:

    -- Application: site_config
    -- Model: WebsiteApplication
    -- Comment: Unknown database type for field 'options' (114)
    ALTER TABLE "site_config_websiteapplication"
    	ALTER "options" TYPE text;

xavfernandez:
This should (partially) fix #1010.

Maybe @imagescape could validate if it fixes it ? (regarding GIS fields).

### DIFF
--- a/django_extensions/management/commands/sqldiff.py
+++ b/django_extensions/management/commands/sqldiff.py
@@ -220,7 +220,7 @@ class SQLDiff(object):
             reverse_type = DATA_TYPES_REVERSE_OVERRIDE[type_code]
         else:
             try:
-                reverse_type = self.introspection.data_types_reverse[type_code]
+                reverse_type = self.introspection.get_field_type(type_code, description)
             except KeyError:
                 reverse_type = self.get_field_db_type_lookup(type_code)
                 if not reverse_type:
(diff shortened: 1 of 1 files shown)


trbs:
LGTM +1

trbs:
Thanks !

PR #1192, opened 2018-04-06
Replace removed `rel` object with `remote_field`

wm3ndez:
Field.rel was removed in Django 1.9:
https://docs.djangoproject.com/en/1.9/releases/1.9/#field-rel-changes


trbs:
We do still support Django 1.8. I know Django's support of it just ended (1-april-2018) but I would like to be a bit more lenient. 

Could you add backwards compatibility code specifically for `Django 1.8` ? (Such that `Django 1.11 LTS` and up is the default case) Then we can remove that in something like 6 months. To give everybody some time to update from Django's end of support.

Thanks !

wm3ndez:
@trbs Done!

PR #1198, opened 2018-04-18
Fix missing special markers `::` in runserver_plus.rst

myii:
* Consistent with the rest of the document
* Without this, `--reloader-interval 5` is rendered with an em dash instead on Read the Docs @ http://django-extensions.readthedocs.io/en/latest/runserver_plus.html#io-calls-and-cpu-usage
    * > –reloader-interval 5


trbs:
Thanks !

myii:
Wow, that was quick!

PR #1203, opened 2018-05-05
add --all option and tests to clear_cache mgmt cmd

### LINKED ISSUE #1202: clear_cache should have an --all option
Hi,

It would make life a bit easier if the `clear_cache` command had an `--all` option, to clear all caches instead of having to name them individually.

Cheers

costela:
does what it says on the title :wink:

fixes #1202

### DIFF
--- a/django_extensions/management/commands/clear_cache.py
+++ b/django_extensions/management/commands/clear_cache.py
@@ -1,8 +1,9 @@
 # -*- coding: utf-8 -*-
 # Author: AxiaCore S.A.S. http://axiacore.com
+from django.conf import settings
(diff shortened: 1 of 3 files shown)


costela:
Not quite sure what's going on with mypy and pypy-dj111 :thinking:
@trbs any ideas off the top of your head?

trbs:
The `pypy-dj111` build is failing a lot recently for no good reason (restarting the build always fixes it). I updated `Travis` to use the newer version of pypy fixes that.

The `mypy` error is also due to it's new version now having strict optional checking, also added a patch for that.

Build should now work :-)

trbs:
Thanks !

PR #1209, opened 2018-06-11
Treat dev versions as unstable in pipchecker

michael-k:
This prevents pipchecker from suggesting upgrading to a dev version when the currently installed version is stable.

See eg. https://pypi.org/project/django-filter/2.0.0.dev1/


trbs:
Thanks !

PR #1216, opened 2018-06-28
Added --nopin option to runserver_plus docs

AndrewGrossman:
The docs for runserver_plus mention using environment variables for disabling the pin, but make no mention of the _nopin_ option that I found in the source and have been using.


trbs:
Thanks !

PR #1217, opened 2018-07-02
Fix autoslug generation when existing model is copied

lskillen:
Fixes the issue described by @tisdall in #1171 that occurs when an existing autoslug model is copied.


trbs:
Thanks !

Issue #1222, opened 2018-07-20
sync_3s fails with python 3.x


radyz:
I've noticed that files are being read in binary mode and StringIO is being used:

- [file_obj](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/sync_s3.py#L372)
- [gzipfile](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/sync_s3.py#L292)

While in python 2.x this worked (StringIO == BytesIO for Django) in python 3.x it breaks.

Do you think an acceptable fix would be to replace that StringIO import to BytesIO? (I've alreay tried it locally with success)

PR #1223, opened 2018-07-20
Add BytesIO buffer for binary content

radyz:
This should solve #1222

### DIFF
--- a/django_extensions/management/commands/sync_s3.py
+++ b/django_extensions/management/commands/sync_s3.py
@@ -66,7 +66,7 @@ from typing import List  # NOQA
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
-from django.utils.six import StringIO
+from django.utils.six import BytesIO
 
 from django_extensions.management.utils import signalcommand
 
@@ -288,7 +288,7 @@ class Command(BaseCommand):
 
     def compress_string(self, s):
         """Gzip a given string."""
-        zbuf = StringIO()
+        zbuf = BytesIO()
         zfile = gzip.GzipFile(mode='wb', compresslevel=6, fileobj=zbuf)
         zfile.write(s)
         zfile.close()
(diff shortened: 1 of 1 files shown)


trbs:
Could you please add some test(s) for this ?

So we know this works on both Py 2.7 and Py 3.x and that we don't regress later :-)

radyz:
Sure, will do later today

trbs:
bump

django_extensions/management/commands/sync_s3.py
```diff
@@ -388,7 +388,7 @@ def upload_s3(self, arg, dirname, names, dirs):
                         print("\tgzipped: %dk to %dk" % (file_size / 1024, len(filedata) / 1024))
             if self.do_expires:
                 # HTTP/1.0
-                headers['Expires'] = '%s GMT' % (email.Utils.formatdate(time.mktime((datetime.datetime.now() + datetime.timedelta(days=365 * 2)).timetuple())))
+                headers['Expires'] = '%s GMT' % (email.utils.formatdate(time.mktime((datetime.datetime.now() + datetime.timedelta(days=365 * 2)).timetuple())))
```
vstoykov: Currently only `email` is imported. In order to work correctly on both versions it should be imported as `import email.utils`. Probably right now it works correctly because somewhere else in the code `email.utils` is imported and now `email` package has `utils` module as attribute attached to it, but I think that it is more safe if `email.utils` is imported as a whole.

django_extensions/management/commands/sync_s3.py
```diff
@@ -66,7 +66,7 @@
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
-from django.utils.six import StringIO
+from django.utils.six import BytesIO
```
vstoykov: I know that Django has `six` module for compatibility but now Django 2.0 is Python 3+ only and in future versions they will probably delete this module. Also `io` module is available as of Python 2.6, and `BytesIO` can be imported from there and will work in Python 2 and 3 (and also is written in C and should be faster than SringIO in Python 2).

I suggest that `from io import BytesIO` should be preferred then `from django.utils.six import BytesIO` for Python 2/3 compatible code which should support Python 2.6+

Issue #1201, opened 2018-04-30
Runserver_plus exceptions stacktrace is polluted


antwan:
Hi there,

When an uncaught exception occurs in my apps using `runserver_plus`, the stacktrace is going through a significant amount of redundant code, and I can't figure out why. 10 calls are made to `six.reraise(exc_type, exc_value, tb)`. I think it should probably be improved for clarity.

Using Django 1.11 / Python 2.7 / Django Extensions last table (2.0.7).

```
127.0.0.1 - - [30/Apr/2018 14:55:01] "GET /crash/ HTTP/1.1" 500 -
Traceback (most recent call last):
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/contrib/staticfiles/handlers.py", line 63, in __call__
    return self.application(environ, start_response)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 157, in __call__
    response = self.get_response(request)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/base.py", line 124, in get_response
    response = self._middleware_chain(request)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 43, in inner
    response = response_for_exception(request, exc)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 93, in response_for_exception
    response = handle_uncaught_exception(request, get_resolver(get_urlconf()), sys.exc_info())
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 139, in handle_uncaught_exception
    return debug.technical_500_response(request, *exc_info)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django_extensions/management/technical_response.py", line 6, in null_technical_500_response
    six.reraise(exc_type, exc_value, tb)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/Users/antwan/Projects/myproject/venv/lib/python2.7/site-packages/django/utils/deprecation.py", line 140, in __call__
    response = self.get_response(request)
[...]

PR #1228, opened 2018-07-31
Reduce reraise pollution in runserver_plus traceback page

jhselvik:
For issue #1201. 

It felt a little hacky but I couldn't figure out a better way to do it, I'm open to suggestions. I tested on 2.7 and 3.7. The output looks much better now.

image

### DIFF
--- a/django_extensions/management/technical_response.py
+++ b/django_extensions/management/technical_response.py
@@ -1,6 +1,31 @@
 # -*- coding: utf-8 -*-
 import six
+from django.core.handlers.wsgi import WSGIHandler
+
+wsgi_tb = None
 
 
 def null_technical_500_response(request, exc_type, exc_value, tb, status_code=500):
-    six.reraise(exc_type, exc_value, tb)
+    """Function to override django.views.debug.technical_500_response.
+
+    Django's convert_exception_to_response wrapper is called on each 'Middleware' object to avoid
+    leaking exceptions. The wrapper eventually calls technical_500_response to create a response for
+    an error view.
+
+    Runserver_plus overrides the django debug view's technical_500_response function with this
+    to allow for an enhanced WSGI debugger view to be displayed. However, because Django calls
+    convert_exception_to_response on each object in the stack of Middleware objects, re-raising an error
+    quickly pollutes the traceback displayed.
+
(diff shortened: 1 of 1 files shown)


trbs:
Should we also clear the global `wsgi_tb` at the end ?

I agree with your comments... feels a bit hacky.. but also I don't have a better answer (or time to see if there is any) at the moment :-)

jhselvik:
That's a good idea, but I'm not sure sure where the end would be. After `null_technical_500_response` re-raises it goes back to the Django code outside of this repo.
 
Each time `null_technical_500_response` is called on Middleware objects it checks to see if it is for a WSGIHandler object. If so it saves the traceback to display. Because of this I believe it should have the correct traceback for each new WSGI request. I tested with consecutive WSGI calls and it work without clearing.

trbs:
Yeah your probably right but still feels kinda "wrong" :-)

Thinking this is very likely also not thread safe. Sure `runserver_plus` should not be used in such a way that this is a real issue (aka in production) but still...

Using thread.local should solve that... (but kinda also feels like a kludge)

As for clearing the global var this would work:
```
    if isinstance(tb.tb_next.tb_frame.f_locals['self'], WSGIHandler):
        wsgi_tb = tb
    elif wsgi_tb:
        tb = wsgi_tb
        wsgi_tb = None
    six.reraise(exc_type, exc_value, tb)
```

Makes it a bit cleaner, clears up the wsgi_tb after using it and removed the duplicate `six.reraise`

jhselvik:
Alright, thanks for taking a look. You're right, I did not consider thread safety.

I am traveling today, but will try and wrap this up with another commit by the end of the weekend. Will clean up the code as mentioned above and look into using thread.local.

jhselvik:
Clearing the global var as mentioned above did not suppress the repeated re-raises in the traceback. `null_technical_500_response` is called on every Middleware object in a stack and after it is called on the WSGI Middleware object the global var was cleared. The WSGI object is called first in the stack, so for the following seven or so `null_technical_500_response` calls the re-raise problem remained because of the global var getting cleared.

However, I removed the global var in favor of threading.local. This is a bit cleaner, but technically still doesn't clear the now thread's wsgi_tb variable. If the threading option is configured each request should have its own thread and is probably not an issue.

trbs:
@jhselvik thanks a lot for all your help on this issue !

Issue #1221, opened 2018-07-19
selector-search.gif missing


bctiemann:
widgets/foreignkey_searchinput.html has the following:

    <img src="{% static "admin/img/selector-search.gif" %}" width="16" height="16" alt="{% trans "Lookup" %}" />

But there is no such image in the statics.

arthurio:
It's a long time know bug (see https://github.com/django-extensions/django-extensions/issues/229 and https://github.com/django-extensions/django-extensions/issues/786).

Since at least Django 1.9 (I don't know for previous versions), the `related-lookup` class will get the image icon from the `Django` css:

```css
.related-lookup {
    width: 16px;
    height: 16px;
    background-image: url(../img/search.svg);
}
```

So I think we can simply drop the `img` tag altogether.

I'd be happy to submit a PR but I'm not sure what are the expectations in terms of compatibility with older versions of Django. Also, I'm sure we can really test the change? Maybe it's not super important since it has not been working for so long?

trbs:
Anything that improves the situation is a step forwards :-)

For Django Extensions we (officially) follow Django in which releases we support, which currently is Django 1.11 and up.

I think what your suggesting is good and happy to merge the PR if you would be so kind to create that.

Issue #786, opened 2015-12-18
ForeignKeyAutocompleteAdmin not showing properly search icon on Dj1.9


danigosa:
It looks for 

```
   <img src="/static/admin/img/selector-search.gif" width="16" height="16" alt="Lookup">
```

While it should be **search.svg**

Screenshot:
<img width="1399" alt="extensions_error" src="https://cloud.githubusercontent.com/assets/1827475/11899552/00a1c372-a5a0-11e5-9d6c-ecce8e0acb6a.png">

arthurio:
I think it doesn't even need an `img` tag since the css on the `a` tag already adds the search icon.

trbs:
Closing this for being inactive for too long. If there is a better way we are still happy to accept a PR for this.

Issue #229, opened 2012-07-31
selector-search.gif wrong path


mdgart:
In /templates/django_extensions/widgets/foreignkey_searchinput.html, line 4:

{{ admin_media_prefix }}img/admin/selector-search.gif

should be:

{{ admin_media_prefix }}img/selector-search.gif

PR #1234, opened 2018-08-07
Remove unnecessary img tag for the foreignkey search input.

arthurio:
This PR fixes https://github.com/django-extensions/django-extensions/issues/1221 (and similar https://github.com/django-extensions/django-extensions/issues/229, https://github.com/django-extensions/django-extensions/issues/786).

As mentioned in the issue, Django's css already add a background image for the `.related-lookup` class, so instead of adding the missing asset `admin/img/selector-search.gif` we can just remove the `img` tag.

Before this PR:

add_registration___ms_kgi_admin

After this PR:

add_registration___ms_kgi_admin


trbs:
Thanks !

PR #1236, opened 2018-08-12
Prevent many2many field with custom intermediate models to be added directly on the parent model

jrd:


jrd:
CI failed because of a timeout and a resource not available on ubuntu at the time the test run.

trbs:
Thanks !

PR #1237, opened 2018-08-14
decode lookup_value if it is bytes instead of str

behconsci:


tests/testapp/models.py
```diff
@@ -82,6 +82,14 @@ class Meta:
         app_label = 'django_extensions'
 
 
+class SluggedTestBytesLookupValueModel(models.Model):
+    title = models.CharField(max_length=42)
+    slug = AutoSlugField(populate_from=b'title')
```
trbs: Why is the `populate_from` attribute specified as bytes `b'title'` here ?

behconsci: because this is the test, right? even if lookup_value is bytes, it should go thru. or what do you think?

trbs: Why would that attribute be every `bytes` though ? 

It reference the model field / lookup path and should always be a string right...

behconsci: yes exactly, but in my case, thru some python version, it was bytes and it failed in .split() method. then I made the fix which checks if `populate_from` is bytes and if yes, just decode and go on.

trbs: I would like to know why that exactly happened and I think the current test does not really do that. The test implies that `bytes` is a valid value type for this attribute and that it should work and the field should convert it. However I'm not entirely sure that the fault in that lies with the Django Extensions field.

`SluggedTestModel` and `test_auto_create_slug` does already test for the normal case and  that passes the build with Python 3.6 and Django 1.11.

So the question is why does it fail for you. 

I'm sorry to not just quickly merge this but it feels like a workaround for me... since this should not happen in the first place.

trbs: I added a check in git to raise `TypeError` when the value passed in `populate_from` is not a callable or a string.

Maybe that will help with a traceback that shows the origin of where the bytes come from...

behconsci: good idea. let me debug with this and will update you here again

behconsci: I am getting this exception message if I apply your change to see the exception: 

```
    goldmarie: Traceback (most recent call last):
    goldmarie:   File "./manage.py", line 10, in <module>
    goldmarie:     execute_from_command_line(sys.argv)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/core/management/__init__.py", line 363, in execute_from_command_line
    goldmarie:     utility.execute()
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/core/management/__init__.py", line 355, in execute
    goldmarie:     self.fetch_command(subcommand).run_from_argv(self.argv)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/core/management/base.py", line 283, in run_from_argv
    goldmarie:     self.execute(*args, **cmd_options)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/core/management/base.py", line 330, in execute
    goldmarie:     output = self.handle(*args, **options)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/core/management/commands/migrate.py", line 83, in handle
    goldmarie:     executor = MigrationExecutor(connection, self.migration_progress_callback)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/db/migrations/executor.py", line 20, in __init__
    goldmarie:     self.loader = MigrationLoader(self.connection)
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/db/migrations/loader.py", line 52, in __init__
    goldmarie:     self.build_graph()
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/db/migrations/loader.py", line 203, in build_graph
    goldmarie:     self.load_disk()
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django/db/migrations/loader.py", line 114, in load_disk
    goldmarie:     migration_module = import_module("%s.%s" % (module_name, migration_name))
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/importlib/__init__.py", line 126, in import_module
    goldmarie:     return _bootstrap._gcd_import(name[level:], package, level)
    goldmarie:   File "<frozen importlib._bootstrap>", line 978, in _gcd_import
    goldmarie:   File "<frozen importlib._bootstrap>", line 961, in _find_and_load
    goldmarie:   File "<frozen importlib._bootstrap>", line 950, in _find_and_load_unlocked
    goldmarie:   File "<frozen importlib._bootstrap>", line 655, in _load_unlocked
    goldmarie:   File "<frozen importlib._bootstrap_external>", line 678, in exec_module
    goldmarie:   File "<frozen importlib._bootstrap>", line 205, in _call_with_frames_removed
    goldmarie:   File "/home/vagrant/goldmarie/goldmarie/multistore/migrations/0001_initial.py", line 10, in <module>
    goldmarie:     class Migration(migrations.Migration):
    goldmarie:   File "/home/vagrant/goldmarie/goldmarie/multistore/migrations/0001_initial.py", line 25, in Migration
    goldmarie:     ('slug', django_extensions.db.fields.AutoSlugField(populate_from=b'title', verbose_name='slug', editable=False, blank=True)),
    goldmarie:   File "/home/vagrant/.virtualenvs/goldmarie/lib/python3.6/site-packages/django_extensions/db/fields/__init__.py", line 116, in __init__
    goldmarie:     raise TypeError("'populate_from' must be str or list[str] found `%s`" % populate_from)
    goldmarie: TypeError: 'populate_from' must be str or list[str] found `b'title'`
```

behconsci: ok wait, please ignore this. there is hand-coded b'title inside the migration. I will come with original exception in a while

PR #1243, opened 2018-08-21
Drop compat code for Django < 1.11 and add classifier for Django 2.1

michael-k:
According to the readme, Django 1.8 is no longer supported, see da5673c43a6e0bc7ebed753c0b66be230f9ae09b and release 2.1.0 was not tested agaist Django 1.8 any more (b6ebd2b9e6cf06195c6241a89d88e5fdf82eb7e3).

This PR goes one step further and removes the compatibility code for Django < 1.11 (mostly it's for Django < 1.9).

Django 1.8 is unsupported for over 4 months now anyway, see https://www.djangoproject.com/download/#supported-versions


trbs:
@michael-k  That was a lot of work to check all the old bits and pieces for old `Django` versions.

Many Thanks !

PR #1244, opened 2018-08-23
show_template_tags should handle AppConfig class in INSTALLED applications

jrd:
https://docs.djangoproject.com/en/2.0/ref/applications/#django.apps.AppConfig

I'm not sure about the `ADMIN_FOR` hack that was done before, but normally `get_app_configs` should return all available applications as an AppConfig class.

My application is listed in `INSTALLED_APPS` with a reference to a `AppConfig` class as it should be possible with Django. This commit correctly show my templates tags and filters.

### DIFF
--- a/django_extensions/management/commands/show_template_tags.py
+++ b/django_extensions/management/commands/show_template_tags.py
@@ -2,7 +2,7 @@
 import inspect
 import os
 
-from django.conf import settings
+from django.apps import apps
 from django.core.management import color
 from django.core.management import BaseCommand
 from django.utils import termcolors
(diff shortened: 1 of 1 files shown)


trbs:
Also not sure what exactly the reasons for `ADMIN_FOR` where... Given the missing documentation for it I don't mind to remove it.

jrd:
Everything works ok using `master`. Thanks for the unit tests ;-)

When do you plan to make a release btw?

trbs:
New release 2.1.1 is uploaded to PyPi :-)

PR #1247, opened 2018-09-02
Fix #1246 regresion in graph models

### LINKED ISSUE #1246: Regression with Python 2 and graph_models
After merging #877 there is regression in `graph_models` command when all of the following conditions are met:

- running on Python 2;
- there are `verbose_name` set for some `ForeignKey`/`ManyToManyFIeld` which contains non ASCII characters;
- `graph_models` is executed with `--verbose-names` argument.

I will try to prepare a PR.

vstoykov:
This fixes #1246 which was caused by #877

### DIFF
--- a/django_extensions/management/modelviz.py
+++ b/django_extensions/management/modelviz.py
@@ -8,8 +8,6 @@ Based on:
   Adapted to be used with django-extensions
 """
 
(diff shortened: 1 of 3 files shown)


trbs:
awesome thanks !

PR #1248, opened 2018-09-04
Fix #1220

### LINKED ISSUE #1220: 2.0.8 breaks ForeignKeyAutocompleteAdmin
I have a model (SocialMediaIdentity) with ForeignKeyAutocompleteAdmin handling the "user" field, which is a ForeignKey to User:

    related_search_fields = {
        'user': ('username', 'email',),
    }

Upgrading from 2.0.7 to 2.0.8 breaks this. Digging through I find that the issue is that model_name has the value "socialmediaidentity" (i.e. the model for which I'm using ForeignKeyAutocompleteAdmin), whereas it should be "user". Thus it looks for username__icontains in the SocialMediaIdentity model instead of in the User model.

arthurio:
See https://github.com/django-extensions/django-extensions/issues/1220#issuecomment-418286073 for details.

I wasn't sure how to test the specific change I made so I created an `admin.py` file in the `testapp` to register the `Person` model and tested the widget behavior against it. Without the fix, the test output would be as follow:

1__arthur_arthurs-macbook-pro____code_django-extensions__zsh_

### DIFF
--- a/django_extensions/admin/widgets.py
+++ b/django_extensions/admin/widgets.py
@@ -35,7 +35,7 @@ class ForeignKeySearchInput(ForeignKeyRawIdWidget):
 
     def label_for_value(self, value):
         key = self.rel.get_related_field().name
-        obj = self.rel.related_model._default_manager.get(**{key: value})
+        obj = self.rel.model._default_manager.get(**{key: value})
 
         return Truncator(obj).words(14, truncate='...')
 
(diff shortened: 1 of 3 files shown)


trbs:
Thanks for the fix and PR !

(I am sorry I don't have the time to track exactly how this came to be. Your PR looks good and thanks for including a test :-)

PR #1250, opened 2018-09-09
Add feature to Graph model : disable_abstract_fields

FloChehab:
[ This PR concerns the Graph Model extension ]

Hi there,

I am currently working on a project where I have a lot of inheritance going on and I was frustrated by the fact that there were no way not to display fields that were inherited.

There were only very few changes to make in order to tackle this issue, and a result, I have added a new command line option "--disable-abstract-fields".

Concerning what I am not sure for this PR : 
- Should I provide some unit tests ? (I am not sure how I would do it)
- Should I suggest a shorter command line attribute ? (with the `-d` taken for  `--disable-fields` I couldn't come up with a none confusing option)

### DIFF
--- a/django_extensions/templates/django_extensions/graph_models/label.dot
+++ b/django_extensions/templates/django_extensions/graph_models/label.dot
@@ -18,11 +18,13 @@
     {{ model.label }}{% if model.abstracts %}<BR/>&lt;<FONT FACE="Helvetica Italic">{{ model.abstracts|join:"," }}</FONT>&gt;{% endif %}
     </FONT></TD></TR>
   {% if not disable_fields %}{% for field in model.fields %}
+  {% if not disable_abstract_fields or not field.abstract %}
     <TR><TD ALIGN="LEFT" BORDER="0">
     <FONT {% if not field.primary_key and field.blank %}COLOR="#7B7B7B" {% endif %}FACE="Helvetica {% if field.abstract %}Italic{% endif %}{% if field.relation or field.primary_key %}Bold{% endif %}">{{ field.label }}</FONT>
     </TD><TD ALIGN="LEFT">
     <FONT {% if not field.primary_key and field.blank %}COLOR="#7B7B7B" {% endif %}FACE="Helvetica {% if field.abstract %}Italic{% endif %}{% if field.relation or field.primary_key %}Bold{% endif %}">{{ field.type }}</FONT>
     </TD></TR>
+  {% endif %}
   {% endfor %}{% endif %}
     </TABLE>
     >]
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -58,6 +58,12 @@ class Command(BaseCommand):
                 'dest': 'disable_fields',
                 'help': 'Do not show the class member fields',
             },
+            '--disable-abstract-fields': {
+                'action': 'store_true',
+                'default': False,
+                'dest': 'disable_abstract_fields',
+                'help': 'Do not show the class member fields that were inherited',
+            },
             '--group-models -g': {
                 'action': 'store_true',
                 'default': False,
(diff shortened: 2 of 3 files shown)


trbs:
Thanks a lot ! 

Yes we need some unit test for this. You can look at some of the other tests for inspiration like `test_show_urls`, `test_export_emails`, etc.

Then create a test which captures the output for the command with and without the option and make sure the lines are respectively in and not-in the output of the command.

django_extensions/templates/django_extensions/graph_models/label.dot
```diff
@@ -18,11 +18,13 @@
     {{ model.label }}{% if model.abstracts %}<BR/>&lt;<FONT FACE="Helvetica Italic">{{ model.abstracts|join:"," }}</FONT>&gt;{% endif %}
     </FONT></TD></TR>
   {% if not disable_fields %}{% for field in model.fields %}
+  {% if not disable_abstract_fields or not field.abstract %}
```
trbs: I really dislike the `double-not-or if` construct. It's a shame Django templates does not support `if not (disable_abstract_fields and field.abstract)`.

Besides it's ugliness, I even prefer:
```
{% if disable_abstract_fields and field.abstract %}
{% else %}
...
{% endif %}
```

To it, just because `double-not-or if` is so hard to read.

FloChehab: I was very close to adding it to the list of unsure points yesterday :100: 
I had in mind the alternative you are suggesting, I will make the change alongside the tests.

trbs: yeah it's a shame there is not really a better alternative as far as I know.. Either it's the `{% else %}` variant or creating a template tag or injecting it into the template context from code. 

Would love for Django templates to support expressions like `{% if not (a and b) %}` :-)

FloChehab: I would need a little help regarding testing. It seems that doing : 
```python
    out = StringIO()
    call_command('graph_models',
                 app_label=['django_extensions'],
                 include_models=['AbstractInheritanceTestModelChild'],
                 disable_abstract_fields=False,
                 stdout=out)
    output = out.getvalue()
```
Doesn't capture the prints from `graph_models.py` (i.e. `output=''`) which is annoying for testing things on those prints.
Should I change the prints by something else ? Or maybe working with files ?

trbs: It should work now.

Added a commit with fixed for this in `graph_models` command and also added some very basic tests to make sure graph_model executes. You can add your tests to `test_graph_models.py` if you want.

FloChehab: Thanks for the help ! 
I have made the modification in the template and added the tests.

FloChehab: Thanks for the help ! I have made the modification in the template and added the tests.

tests/management/commands/test_graph_models.py
```diff
@@ -27,3 +27,27 @@ def test_graph_models_json():
...
+
+def test_disable_abstract_fields_not_active():
+    out = StringIO()
+    call_command('graph_models',
+                 app_label=['django_extensions'],
+                 include_models=['AbstractInheritanceTestModelChild'],
+                 disable_abstract_fields=False,
+                 stdout=out)
+    output = out.getvalue()
+    print(output)
```
trbs: please remove the debug print statement :-)

trbs:
awesome thanks !

tests/management/commands/test_graph_models.py
FloChehab: Oups, I went to fast sorry

FloChehab: It's corrected

Issue #1251, opened 2018-09-10
2.0.8: AutoSlugField: Given slug in object creation gets ignored


MerlinB:
When creating a TitleSlugDescriptionModel with title and slug, preference if given to the autoslug generation from title, and any given slug is ignored.
In 2.0.7 calling  `Model.objects.create(slug='key', title='title')` would create an object with `slug = 'key'`
Since 2.0.8 the object gets created with `slug='title'`.
As this change is not documented in the changelog I assume it is unwanted.

trbs:
Yeah seems to be an unwanted side effect.

Could you please make a PR with a test case for this ? to avoid and catch future regressions.

trbs:
Reason for the change is here I think: https://github.com/django-extensions/django-extensions/commit/3eaa21c694a45e8d5f427ccb1c124dfaa4b67b1c#diff-d3086342951f62e09f62d2c4189097d2R146

In quickly looking at the problem I don't have a solution right away.

That case is about when a models gets copied then the slug should be recalculated. Where as here when the slug is explicitly given it should be used as is...

Not sure if we can solve this without basicaly changing the code back again to what it was and saying that in the case of copying the model one should not only clear `pk` but also the `slug` field manually...

tisdall:
Here's the thing...  The change made was to return to how things were working prior to 2.0.7 .  2.0.7 is the only version where the example given actually works.  That "feature" was an unintentional breaking change.

If I were to be designing the field from scratch, then I think only setting the slug if it's currently blank would be a good idea.  However, changing the existing functionality will likely break some people's code who aren't expecting this change (it would break my code, just as it did in 2.0.7).  I'd suggest adding another argument to `AutoSlugField` to explicitly turn on the option to only generate the slug when it's blank.  Or, I'd provide instructions in the docs on how to add that functionality by extending the `AutoSlugField`.  Or, perhaps provide an already extended `AutoSlugField` with a different name.  Or increment the version to indicate a breaking change (which would be simpler if this wasn't a collection of so many extensions).

Here's an example implementation:
```python
class MostlyAutoSlugField(AutoSlugField):
    def create_slug(self, model_instance, add):
        if getattr(model_instance, self.attname):
            # force not overwriting slug if already set
            add = False
        return super(MostlyAutoSlugField, self).create_slug(model_instance, add)
```

trbs:
I'm good with the solution to add a option for this behavior.

[...]

PR #1252, opened 2018-09-10
Add test for slug argument priority in model creation

MerlinB:
Related Issue: https://github.com/django-extensions/django-extensions/issues/1251


trbs:
Thanks !

PR #1253, opened 2018-09-18
Adds ability to enforce_max_length

jamespeacock:
Adds a property to BaseEncryptedField 'enforce_max_length' that defaults to False.
Upon encrypting, the library currently warns and truncates the field value if it is beyond max length.
These changes allow for an error to be raised instead if enforce_max_length is set to true.

A test has also been created to verify the error is raised.

### DIFF
--- a/django_extensions/db/fields/encrypted.py
+++ b/django_extensions/db/fields/encrypted.py
@@ -6,6 +6,7 @@ from django import forms
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db import models
+import logging
 
 try:
     from keyczar import keyczar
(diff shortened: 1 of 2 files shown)


django_extensions/db/fields/encrypted.py
```diff
@@ -6,6 +6,7 @@
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
 from django.db import models
+import logging
```
trbs: unused import ?

jamespeacock: Good catch, was using this to debug. Have some more changes coming in as well. Will push them together

PR #1254, opened 2018-09-24
Add direct link to screencast video

glasnt:
The original blog post links to a direct media download from the blog itself, which no longer works. A quick search shows that https://vimeo.com/1720508 should be the same screencast, but externally hosted. (The Vimeo link also links back to the original post)


trbs:
Thanks @

PR #1255, opened 2018-09-25
Improve module existence test in runscript

jscissr:
Check if the import system actually didn't find the module, instead of just guessing the file name and testing if that file exists.
Because find_spec was added in Python 3.4, we still keep the old method.

I had the issue that import errors inside my scripts were swallowed because the scripts and manage.py are not where BASE_DIR is. That is fixed with this PR.

If you think this needs a test, tell me.


trbs:
Please add tests, that would be awesome !

jscissr:
I have added a test, please review this PR

trbs:
Thanks !

Issue #1204, opened 2018-05-17
graph_models "-o output.dot" should work without pygraphviz or pydot


riconnon:
I'm trying to get the raw dot output from `graph_models` in a constrained environment without pygraphviz or pydot (application docker image)

It seems like the -o flag should work for raw dot output without these dependencies.

trbs:
Very happy to accept a PR for this ! 👍

PR #1257, opened 2018-10-06
adding default output to graph_models

rjauquet:
Hello. Gave a go at addressing #1204 -- if I'm understanding correctly we just want to output the dotdata directly when neither `pygraphviz` or `pydot` are present.

### DIFF
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -198,7 +198,7 @@ class Command(BaseCommand):
             elif use_pydot:
                 self.render_output_pydot(dotdata, **options)
             else:
-                raise CommandError("Neither pygraphviz nor pydotplus could be found to generate the image")
+                self.render_output_default(dotdata, **options)
         else:
             self.print_output(dotdata)
 
@@ -259,3 +259,13 @@ class Command(BaseCommand):
         ext = output_file[output_file.rfind('.') + 1:]
         format = ext if ext in formats else 'raw'
         graph.write(output_file, format=format)
+
+    def render_output_default(self, dotdata, **kwargs):
+        """Renders raw dot image"""
+        if six.PY3 and isinstance(dotdata, six.binary_type):
+            dotdata = dotdata.decode()
+
+        output_file = kwargs['outputfile']
+
+        with open(output_file, 'w') as ofile:
+            ofile.write(dotdata)


trbs:
Thanks !

In the documentation we state that the same functionality is also available without using `-o` at all. https://github.com/django-extensions/django-extensions/blob/master/docs/graph_models.rst

What I would suggest is to add a `--dot` option (or similar) to explicitly write out dotfile data, much like the `--pydot` and `--pygraphviz` options. 

Then we should decide if we only want to write out dot data when `--dot` is specificed with `-o` or automatically such as per the PR although that does break backwards compatibility somewhat if your expecting `graph_models` to fail right now if you do not have the right dependencies installed. We could document an backwards compatiblity break for later releases.

Lastly please update the documentation accordingly and lets add some tests to cover these cases.

Thanks !

jonemo:
I was fishing around the issues list for another reason but noticed #1204 which I ran into just yesterday. Feel free to ignore my drive-by PR comment, but since it's fresh on my mind I'm going to add my 2 cent here:

Based on the output of `manage.py graph_models --help` I would have expected yet another behavior than the ones described in the discussion above:

```
$ python manage.py graph_models --help
usage: manage.py graph_models [-h] [--version] [-v {0,1,2,3}]
                              [--settings SETTINGS] [--pythonpath PYTHONPATH]
                              [--traceback] [--no-color] [--pygraphviz]
                              [--pydot] [--disable-fields] [--group-models]
                              [--all-applications] [--output OUTPUTFILE]
                              [--layout LAYOUT] [--verbose-names]
                              [--language LANGUAGE]
                              [--exclude-columns EXCLUDE_COLUMNS]
                              [--exclude-models EXCLUDE_MODELS]
                              [--include-models INCLUDE_MODELS]
                              [--inheritance] [--no-inheritance]
                              [--hide-relations-from-fields]
                              [--disable-sort-fields] [--json]
                              [app_label [app_label ...]]

Creates a GraphViz dot file for the specified app names. You can pass multiple
app names and they will all be combined into a single model. Output is usually
directed to a dot file.

positional arguments:
  app_label

optional arguments:
[...]
  --pygraphviz          Use PyGraphViz to generate the image.
  --pydot               Use PyDot(Plus) to generate the image.
[...]
  --output OUTPUTFILE, -o OUTPUTFILE
                        Render output file. Type of output dependend on file
                        extensions. Use png or jpg to render graph to image.
[...]
  --json                Output graph data as JSON
```

> Render output file. Type of output dependend on file extensions. Use png or jpg to render graph to image.

This suggests that the output format is chosen based on file extension. I would have expected that `-o myfile.dot` works without pygraphviz or pydot and writes a .dot file (same as #1204). 

Similarly, I would expect that `-o myfile.json` works without pygraphiviz or pydot. It does not, but `--json -o myfile.json` does.

Two possible conclusions:
* Introduce `--dot` flag as suggested by @trbs. Clarify help output for `-o` option to no longer state "Type of output dependend on file extensions"
* Keep "Type of output dependend on file extensions" help text and handle `.json` and `.dot` file extensions separately, dropping need for `--json` (but keeping it for backward compatibility).

trbs:
Thanks @jonemo 

I really like automatically choosing the format based on the extension if no other cli option is given.

And why not do both :-) if no other option is given base the format on the extension of the file. I can imaging certain workflows where people do `./manage.py graph_models -a --json --output /some/temp.file` which might not necessarily then end with `temp_file.json`.

@rjauquet or @jonemo are you able to update the PR ?

jonemo:
> are you able to update the PR ?

@trbs: I'll open a PR within the next days. I have something working but need to familiarize myself with how testing works in this project :)

PR #1258, opened 2018-10-12
Drop pyparsing version in documentation as it breaks on py3

tino:
Using without a version makes it work on py3


trbs:
Thanks !

Not entirely sure but I do remember at the time there was a need for that specific version.
Cannot remember why though :-) Next time we should add a note to the docs about why that specific version.

PR #1259, opened 2018-10-16
Improved output options for graph_models command

jonemo:
Addresses #1204 and implements a suggestion discussed in #1257.

This refactors the logic for deciding which output format to generate in the `graph_models` management command to add three new behaviors:
* Using the `--output` option with a filename ending in `.dot` is now possible when neither `pydot` nor `pygraphviz` are installed (#1204)
* There is a new `--dot` option to explicitly request dotfile output, for example when writing to an output file with no file extension: `--dot -o filename`.
* When the `--output` option is used with filenames ending `.dot` or `.json`, the output format is inferred from the file extension unless overridden by one of the output format options (`--json`, `--dot`, `--pydot`, `--pygraphviz`).

To support the changes, the help text and inline comments have been slightly adjusted and some unit test coverage has been added.

### DIFF
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -37,7 +37,8 @@ class Command(BaseCommand):
         space-separated args and the value is our kwarg dict.
 
         The default from settings is keyed as the long arg name with '--'
-        removed and any '-' replaced by '_'.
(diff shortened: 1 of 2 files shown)


trbs:
Thanks !

Issue #1229, opened 2018-07-31
Excluded models displayed as an underscore


donkirkby:
Thanks for building the `graph_models` command, we find it very useful for our documentation. Unfortunately, we noticed a problem after we upgraded to a more recent version. I expected that excluding a model from `graph_models` would completely hide it from the graph. Instead, all my hidden models appear as an underscore when they are referenced by visible models.

Here is a simple example:

    # container/models.py
    from django.db import models
    
    
    class Foo(models.Model):
        name = models.CharField(max_length=30)
    
    
    class Bar(models.Model):
        name = models.CharField(max_length=30)
        foo = models.ForeignKey(Foo)

I graph that with this command:

    ./manage.py graph_models --pygraphviz container -o container.png

That produces the expected graph:

![good container]

When I exclude the `Foo` model with this command:

    ./manage.py graph_models --pygraphviz container -o container.png --exclude-models Foo

The model is replaced by an underscore. If the model is referenced by many other models, they are all linking back to this underscore.

![underscore container]

Is there a workaround for this? My current plan is to edit the dot file after it's generated, but I'd appreciate an easier alternative.

I tested this with django extensions 2.1.0.

[good container]: https://user-images.githubusercontent.com/1639148/43482308-2d2ab91c-94bd-11e8-8055-334cb9d0d54e.png
[underscore container]: https://user-images.githubusercontent.com/1639148/43482448-929630ba-94bd-11e8-9e9e-a0d21f9b191f.png

trbs:
Pretty sure that this is a bug :-) Thanks for reporting it !

PR #1266, opened 2018-11-07
Fix for bug #1229.

donkirkby:
I reported issue #1229, and then it got the `need test case` label. This pull request contains a test case, but not a fix.

Hope that's helpful.


donkirkby:
OK, I found a fix. An alternative would be to make `ModelGraph.add_relation()` actually add the relation instead of returning it, but this seemed like a smaller change.

trbs:
awesome !

PR #1267, opened 2018-11-09
Update graph_model css to match 2018 style djangoproject.com

glasnt:
First cut of an attempt to update the default style from the ~2005 version of djangoproject to the current style.

Removes pale golden rod and olive green for djangoproject green and white. Also uses Roboto, which is a Google Font, but for some output formats would be required to be installed natively on the system (as would Helvetica Bold, which is not on maOS Mojave, causing some font issues when generating SVGs)

Sample before/after

screen shot 2018-11-10 at 9 32 54 am

### DIFF
--- a/django_extensions/templates/django_extensions/graph_models/digraph.dot
+++ b/django_extensions/templates/django_extensions/graph_models/digraph.dot
@@ -3,18 +3,18 @@
   // Created: {{ created_at }}
   {% if cli_options %}// Cli Options: {{ cli_options }}{% endif %}
 
-  {% block digraph_options %}fontname = "Helvetica"
+  {% block digraph_options %}fontname = "Roboto"
   fontsize = 8
   splines  = true{% endblock %}
 
   node [{% block node_options %}
-    fontname = "Helvetica"
+    fontname = "Roboto"
     fontsize = 8
     shape = "plaintext"
   {% endblock %}]
 
   edge [{% block edge_options %}
-    fontname = "Helvetica"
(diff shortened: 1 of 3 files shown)


trbs:
Thanks for the PR !

Sorry for the late response, I would like to merge this but we need to fix the tests.

Also I think we should add `theming` options, so people can still use the old version if they wish so..

glasnt:
@trbs Which style should be the default, this new one, or the old one? 

There'd be a few widgets to change in order to incorporate multiple themes, but it shouldn't be too hard to extend.

trbs:
> @trbs Which style should be the default, this new one, or the old one?
> 
> There'd be a few widgets to change in order to incorporate multiple themes, but it shouldn't be too hard to extend.

Would be nice to keep the old as the default for couple of releases then switch... but I don't mind switching over directly as well, specially since it's the style on `django` now already.

Awesome if you want to help with multiple themes as well. I will help as well as time permits !

glasnt:
I'm sorry I haven't had time to look at this, I know it's not the most important thing in the world, but if you're expecting a completed PR from me, I'd greatly appreciate a link to an example of the sort of switching you'd expect here (a sample from elsewhere in django-extensions that's your best practice). Thanks

trbs:
I was thinking of putting the files into separate 'theme' directories. 

I committed a solution to this in master. Let me know what you think !
Thanks !

glasnt:
Works well! Nice work!

PR #1271, opened 2018-11-15
Update set_fake_passwords.py

arrahimi2:
requires_model_validation has been replaced with requires_system_checks sinc 1.9 see https://docs.djangoproject.com/en/1.11/internals/deprecation/

### DIFF
--- a/django_extensions/management/commands/set_fake_passwords.py
+++ b/django_extensions/management/commands/set_fake_passwords.py
@@ -18,7 +18,7 @@ DEFAULT_FAKE_PASSWORD = 'password'
 
 class Command(BaseCommand):
     help = 'DEBUG only: sets all user passwords to a common value ("%s" by default)' % (DEFAULT_FAKE_PASSWORD, )
-    requires_model_validation = False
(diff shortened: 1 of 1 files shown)


trbs:
Thanks !

Do you maybe want to write a PR with some (simple) tests as well ? so we can catch these types of errors next time in the test suite :-)

arrahimi2:
Your welcome, not sure if that is possible maybe fail when you get any deprecation warnings?
Also I am upgrading our software to django 1.11 and we use your extension, I will create pr's if I see more things.

If it's possible to make automated test it would be indeed very good and we'd like to use this. However if it's not it's good to check these pages before upgrading the django version [docs](https://docs.djangoproject.com/en/1.11/howto/upgrade-version/#dependencies)

trbs:
Thanks for using `django-extensions` and helping out !

I'm trying to get to a point where we at least have simple test for each command :-)
That should also cover when we miss a deprecation like `requires_model_validation`.

Will try to add some tests and do a release today.

PR #1277, opened 2018-12-14
Missing schedule documentation

gtmsingh:
Added guide for schedules supported by library but not directly available using `python manage.py create_jobs <app name>` command


trbs:
Thanks !

PR #1286, opened 2019-01-14
Tests for reset_db management command

kuter:


kuter:
I've created tests for reset_db command .. all tests passed but there is one thing that I would change is catching stdout now I using mock .. according to Django documenation it should be call_command('cmd', stdout=out) which won't work.

trbs:
Love more tests 👍  Thanks !

PR #1288, opened 2019-01-19
Tests for management command syncdata

kuter:


kuter:
I've added tests and docs to syncdata management command. I've also made small fixes.

trbs:
Thanks !

PR #1296, opened 2019-01-31
Tests for set_default_site command

kuter:
- tests for set_default_site management command added
- small improvements in that command


trbs:
Thanks !

PR #1297, opened 2019-02-03
Tests for drop_test_database management command

kuter:


django_extensions/management/commands/drop_test_database.py
```diff
@@ -41,10 +41,6 @@ def handle(self, *args, **options):
         """
         Drop test database for this project.
         """
-
-        if args:
-            raise CommandError("reset_db takes no arguments")
```
trbs: This removal changes behavior, is there a reason for it ? (which cannot be solved otherwise)

trbs:
Awesome ! I love more tests !

django_extensions/management/commands/drop_test_database.py
```diff
@@ -136,8 +132,9 @@ def handle(self, *args, **options):
 
             try:
                 cursor.execute(drop_query)
-            except Database.ProgrammingError as e:
+            except (Database.ProgrammingError, Exception) as e:
```
trbs: What's the reason for catching the very wide `Exception` ?
This makes `Database.ProgrammingError` not necessary anymore I think...
I'd rather not catch `Exception` if it's not absolutely necessary...

kuter: ups, I put that general exception only for test purpose and I forget to remove it.

django_extensions/management/commands/drop_test_database.py
kuter: I actually have no idea how can I pass some arguments using management command so I think that exception will never happen

trbs: Your probably right.. might be a left over from the old way of parsing command options in Django Commands...

django_extensions/management/commands/drop_test_database.py
trbs: Please commit and push and I will merge 👍

trbs:
Thanks @kuter !

kuter:
@trbs you're welcome ;)

Issue #1230, opened 2018-08-02
SyntaxError: invalid syntax in dumpscript


martinvol:
After generating a dumpscrip, I called it somewhere else and I got a syntax error. I'm aware this is not the way the documentation suggest for running the script, but the syntax error shouldn't be there anyway, I didn't try but it would also fail doing `./manage.py runscript out`

`python out.py`

```
  File "out.txt", line 136
    pictures_picture_3.owner_user =  importer.locate_object(User, "id", User, "id", 1, {'id': 1, 'password': 'XXXX', 'last_login': datetime.datetime(2018, 7, 30, 3, 19, 37, 885214, tzinfo=<UTC>), 'is_superuser': True, 'first_name': 'Martín', 'last_name': 'Volpe', 'is_staff': True, 'is_active': True, 'date_joined': datetime.datetime(2010, 1, 11, 21, 45, 36, tzinfo=<UTC>), 'email': 'XX@gmail.com', 'slug': 'martinvol'} )
                                                                                                                                                   ^
SyntaxError: invalid syntax

```

The error is with `tzinfo=<UTC>`

Django 2.0.7
django-extensions 2.0.7
Pyton 3.6.4

If you are struggling with the error yourself, you can walk around generating the user manually (if it doesn't exist yet) and replace 

`owner_user =  importer.locate_object(User, "id", User, "id", 1, {'id': 1, 'password': 'XXXX', 'last_login': datetime.datetime(2018, 7, 30, 3, 19, 37, 885214, tzinfo=<UTC>), 'is_superuser': True, 'first_name': 'Martín', 'last_name': 'Volpe', 'is_staff': True, 'is_active': True, 'date_joined': datetime.datetime(2010, 1, 11, 21, 45, 36, tzinfo=<UTC>), 'email': 'XX@gmail.com', 'slug': 'martinvol'} ) `

with:

`owner_user =  User.objects.get(email='XX@gmail.com')`

Best,

trbs:
Seems an error in serializing those datetime's. In other places it's serialized as:
```
"dateutil.parser.parse(\"%s\")" % value.isoformat()
```

This is probably the best way to do it here as well.

You think you can create a PR for this or at least a PR that creates a test in the testsuite that fails on the current behavior ? (eg: asserting that the output does not contain <UTC>)

the-bionic:
Happy to pick this up as a new contributor.

trbs:
[...]

PR #1298, opened 2019-02-04
Fix orm_item_locator to use dateutil

yeojin-dev:
Issue #1230 
I think the issue is caused by `tzinfo=<UTC>`. Because `<UTC>` isn't UTC, tzinfo instance.
In case of ForeignKey and M2MField, it's ok to make datetime as string.


trbs:
Why not implement it as: `"dateutil.parser.parse(\"%s\")" % value.isoformat()` such is done in `get_attribute_value(...)` ?

yeojin-dev:
Sorry. I misunderstood. Could you see my new commit? @trbs

trbs:
Thanks @yeojin-dev !

One last question; could you amend your last patch to use:

```
    clean_dict[key] = "dateutil.parser.parse(\"%s\")" % v.isoformat()
```

As I think this would fail for timezone aware datetimes. Those would look like `datetime.datetime(2003, 10, 11, 17, 13, 46, tzinfo=tzoffset(None, 7200))` and the `tzinfo` items like `tzoffset`, `tzlocal`, `tzutc`, etc are not imported.

yeojin-dev:
Thank you for reviewing! I've fixed it.

yeojin-dev:
Thank you for reviewing! I've fixed it. @trbs

trbs:
Thanks @yeojin-dev for all your work on this PR. 

I hate to say it but i would prefer to merge the other solution from my comment. As this is the same as we use elsewhere in the command and does not change the objects timezone.

django_extensions/management/commands/dumpscript.py
```diff
@@ -72,6 +73,7 @@ def orm_item_locator(orm_obj):
         v = clean_dict[key]
         if v is not None:
             if isinstance(v, datetime.datetime):
+                v = v.replace(tzinfo=get_current_timezone())  # get tzinfo from settings.py
```
MRigal: I would actually rather use:
```
from django.utils.timezone import make_aware
v = timezone.make_aware(v)
clean_dict[key] = dateutil.parser.parse(v.isoformat())
```

django_extensions/management/commands/dumpscript.py
```diff
@@ -70,8 +71,12 @@ def orm_item_locator(orm_obj):
 
     for key in clean_dict:
         v = clean_dict[key]
-        if v is not None and not isinstance(v, (six.string_types, six.integer_types, float, datetime.datetime)):
-            clean_dict[key] = six.u("%s" % v)
+        if v is not None:
+            if isinstance(v, datetime.datetime):
+                v = v.replace(tzinfo=get_current_timezone())  # get tzinfo from settings.py
+                clean_dict[key] = dateutil.parser.parse(v.isoformat())
```
MRigal: I still wonder why you need to reparse it... Why not directly `clean_dict[key] = v.isoformat()`?

trbs: It's been a while since I looked at this but indeed there is no reason to reparse it here.

The idea was to do exactly the same thing as elsewhere in the `dumpscript` code. That is have the resulting script do the parsing back from a isoformat to a python datetime object.

Now it's using `clean_dict[key] = dateutil.parser.parse(v.isoformat())` which results in a string like this in the dumpscript for ` datetime.datetime.now().replace(tzinfo=pytz.utc)`:

```python
importer.locate_object(orgklass, "orgid", klass, "pk_name", 123,  {'test': datetime.datetime(2019, 4, 28, 22, 24, 1, 827687, tzinfo=tzutc())})
```

Which I think will raise an `NameError` since `tzutc` is never imported only `tzoffset` is.

What we want is something that looks like this:
```
importer.locate_object(orgklass, "orgid", klass, "pk_name", 123,  {'test': dateutil.parser.parse("2019-04-28T22:26:57.248346+00:00")}
```

Then `locate_object` gets a datetime object instead of just the string with isoformat time in it.

That said... I'm not sure if (and if so how much) this is actually used in the wild.. and the isoformatted string might be 'good enough' if nobody is exactly using the `obj_content` for `locate_object`.

django_extensions/management/commands/dumpscript.py
yeojin-dev: Thank you!

django_extensions/management/commands/dumpscript.py
yeojin-dev: I misunderstood. :sweat_smile:

yeojin-dev: I misunderstood. 😅 @trbs

trbs:
Thanks @yeojin-dev could you elaborate a little bit on why the `datetimes_to_replace` list is needed ? (from first glance it's not really obvious)

yeojin-dev:
Codes may behave incorrectly when there are multiple datetime fields. It is necessary for a datetime to be changed exactly one by one into a dateutil.

tests/testapp/settings.py
```diff
@@ -84,3 +84,5 @@
         'BACKEND': 'tests.management.commands.test_clear_cache.OtherCacheMock',
     },
 }
+
+TIME_ZONE = 'Asia/Seoul'
```
trbs: Better to set the `TIME_ZONE` explicitly on the tests then to force all the tests to run on the `Asia/Seoul` timezone.

Something like: https://pytest-django.readthedocs.io/en/latest/helpers.html#settings might help

trbs:
> Codes may behave incorrectly when there are multiple datetime fields. It is necessary for a datetime to be changed exactly one by one into a dateutil.

I am sorry, I still do not really understand why this is needed. Also using `number + space + isoformat` placeholder and then doing a big search + replace over the output does not feel right..

I would like to understand why this is the case and if we cannot solve it in a more elegant way.

yeojin-dev:
Sorry for the late reply.
It is necessary to avoid this. dateutil is a string type, not a code.
```
django_extensions_note_1.user =  importer.locate_object(
    ... 'date_joined': 'dateutil.parser.parse("2019-06-15T10:17:18.931374-05:00")'} )
```

```
    for key in clean_dict:
        v = clean_dict[key]
        if v is not None:
            if isinstance(v, datetime.datetime):
                v = timezone.make_aware(v)
                clean_dict[key] = 'dateutil.parser.parse("%s")' % v.isoformat()
            elif not isinstance(v, (six.string_types, six.integer_types, float)):
                clean_dict[key] = six.u("%s" % v)

    output = """ importer.locate_object(%s, "%s", %s, "%s", %s, %s ) """ % (
        original_class, original_pk_name,
        the_class, pk_name, pk_value, clean_dict
    )
    return output
```
In the above code, datetime can only enter a string. It's hard to do this without changing output. So I used `replace()`. I need an idea of how to enter datetime as a code instead of string. @trbs

tests/testapp/settings.py
yeojin-dev: I fixed it.

yeojin-dev:
I fixed it. :) @trbs

yeojin-dev:
Sorry for the late reply. I fixed it. Would you please review my code? @trbs

trbs:
@yeojin-dev thanks for all your hard work on this !

PR looks good to me :-)

PR #1300, opened 2019-02-04
Tests for management command sqldsn

kuter:
- added acceptance tests for command
- sqldsn command code refactoring (you can run tests against old code - all tests pass)

### DIFF
--- a/django_extensions/management/commands/sqldsn.py
+++ b/django_extensions/management/commands/sqldsn.py
@@ -73,22 +73,12 @@ class Command(BaseCommand):
         dsn = []
 
         if engine == 'mysql':
-            dsnstr = 'host="{0}", db="{2}", user="{3}", passwd="{4}"'
-            if dbport is not None:
-                dsnstr += ', port="{1}"'
-
-            dsn.append(dsnstr.format(dbhost,
-                                     dbport,
-                                     dbname,
-                                     dbuser,
-                                     dbpass))
-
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/sqldsn.py
```diff
@@ -129,23 +127,14 @@ def postgresql(self, dbhost, dbport, dbname, dbuser, dbpass, dsn_style=None):
...
-                dsnstr = "postgresql://{3}:{4}@{0}/{2}"
+            dsnstr = "postgresql://{user}:{password}@{host}/{name}"
 
-            dsn.append(dsnstr.format(dbhost,
-                                     dbport,
-                                     dbname,
-                                     dbuser,
-                                     dbpass,))
+            dsn.append(dsnstr.format(
+                host=":".join(map(str, filter(None, [dbhost, dbport]))),
```
trbs: Could we use: `"{host}:{port}".format(host=dbhost, port=dbport) if dbport else dbhost`
(shorter `"%s:%s" % (dbhost, dbport)` is also fine) instead of the `map` and `filter` combo ?

Which I like for data parsing but imho makes these cases harder to read.

kuter: @trbs changed, readability counts ;)

trbs:
awesome thanks !

`#noqa` out of interest what was it complaining about there ?

kuter:
> awesome thanks !
> 
> `#noqa` out of interest what was it complaining about there ?

line is too long .. so I put noqa ;)

PR #1306, opened 2019-02-13
Fix and tests for management command print_user_for_session

kuter:


kuter:
All tests passed but it looks like there was a problem with AWS S3 so it cannot execute on py35 and dj20

trbs:
restarted that build, should be green soon :-)

kuter:
great, if it's ok please merge this PR and others that are waiting ;)

trbs:
Thanks !

PR #1309, opened 2019-02-16
removed passwd management command

kuter:


trbs:
Could you explain the reason for this PR please ?

kuter:
@trbs I believe that this command was useful few years ago .. nowadays we've got build-in changepassword command.

trbs:
I understand... I would like to deprecate the command first for couple of releases instead of outright removing it. (Steps should probably be... print warning for couple of releases with functioning command, then only print warning and exit, then remove completely)

kuter:
I've created new PR #1331 which add deprecation warning and print that warning for passwd command.

trbs:
Awesome ! I merged that and build failures should stop soon as well :-)

PR #1315, opened 2019-02-20
Tests for indentby tag

kuter:


django_extensions/templatetags/indent_text.py
```diff
@@ -45,7 +45,8 @@ def indentby(parser, token):
     args = token.split_contents()
     largs = len(args)
     if largs not in (2, 4):
-        raise template.TemplateSyntaxError("%r tag requires 1 or 3 arguments")
+        raise template.TemplateSyntaxError(
```
trbs: Could we keep the old version of the intentation please ? :-)
(We have `max-line-length = 512` in `setup.cfg`)

kuter: sure we can .. how about change %r into indentby ? tests gives me %r which is not very informative

trbs: Think the `%r` is a typo and fully support the change itself 👍

trbs:
awesome ! thanks for all your help !

PR #1316, opened 2019-02-20
Additional tests for encrypted fields

kuter:


tests/test_encrypted_fields.py
```diff
@@ -4,15 +4,19 @@
...
+from django.test.utils import override_settings
 
 from .testapp.models import Secret
 
 # Only perform encrypted fields tests if keyczar is present. Resolves
 # http://github.com/django-extensions/django-extensions/issues/#issue/17
 try:
-    from django_extensions.db.fields.encrypted import EncryptedTextField, EncryptedCharField  # NOQA
+    from django_extensions.db.fields.encrypted import BaseEncryptedField, \
+        EncryptedTextField, EncryptedCharField
```
trbs: Can we either put all the imports on one line or make multiple import lines instead of using a `\` ?

trbs:
Otherwise LGTM +1 👍

tests/test_encrypted_fields.py
kuter: I've put imports in one line. What style of multiple lines would you prefrer (I will kept that in my memory for the furure) ?

kuter:
I've made all changes that you request, if its OK merge that please ;)

tests/test_encrypted_fields.py
trbs: I actually prefer to just use a single line until it becomes too unwieldy.
This one is still just 108 chars wide.. so good enough for me to keep on one line...

Personally I would break pretty late ... maybe around 200 chars or so.. 

What I would actually prefer but isort does not support is breaking like this:
```
from third_party import lib1, lib2, lib3
from third_party import lib4, lib5
```

( I have added `line_length = 128` to isort config in setup.cfg. )

For now I would just keep this import on one line.

kuter: ok, its in one line now.

tests/test_encrypted_fields.py
```diff
@@ -240,3 +243,57 @@ def test_encrypt_public_decrypt_private(self):
...
+    @classmethod
+    def setUpClass(cls):  # noqa
+        cls.tmpdir = tempfile.mkdtemp()
+        keyczart.Create(cls.tmpdir, "test", keyinfo.DECRYPT_AND_ENCRYPT,
+                        asymmetric=True)
+        keyczart.AddKey(cls.tmpdir, "PRIMARY", size=4096)
+
+    @classmethod
+    def tearDownClass(cls):  # noqa
+        import shutil
```
trbs: One last question :-) any reason to do the import here instead of on the top of the file ?

kuter: there was no reaston - it was just copy&paste ;) you're right so I've moved that import on the top of the test file.

trbs: 👍 
When the commit hits the branch / pr I will merge it.
Thanks as always for your work !

kuter: you're welcome, the goal is to reach green coverage bandage .. but it starts to be difficult ;)

Issue #1295, opened 2019-01-31
Django 2.2 compatibility


douglasmiranda:
When running `runserver_plus` with django==2.2a1 we'll get this error:

```
# ...
File "/django/.local/lib/python3.7/site-packages/django_extensions/management/commands/runserver_plus.py", line 17, in <module>
    from django.utils.autoreload import gen_filenames
ImportError: cannot import name 'gen_filenames' from 'django.utils.autoreload' (/django/.local/lib/python3.7/site-packages/django/utils/autoreload.py)
~ $ 
```

This is the commit where they remove `gen_filenames`: 
- https://github.com/django/django/commit/c8720e7696ca41f3262d5369365cc1bd72a216ca#diff-46e69f287173eef41fcbfeba05501954

I haven't tested yet but maybe `BaseReloader.watched_files` could do the trick.
- https://github.com/django/django/blob/2.2a1/django/utils/autoreload.py#L235

trbs:
Could you please test with 073ac63a1def6099fb6a3d3c5717717086a61ad4 ?

douglasmiranda:
Yep, that seems to fix the issue. xD 

But, we'll have an issue when using `USE_I18N = True`.

```
File "/django/.local/lib/python3.7/site-packages/django_extensions/management/utils.py", line 59, in inner
  ret = func(self, *args, **kwargs)
File "/django/.local/lib/python3.7/site-packages/django_extensions/management/commands/runserver_plus.py", line 260, in handle
  self.inner_run(options)
File "/django/.local/lib/python3.7/site-packages/django_extensions/management/commands/runserver_plus.py", line 360, in inner_run
  extra_files.extend(filter(lambda filename: filename.endswith('.mo'), gen_filenames()))
File "/django/.local/lib/python3.7/site-packages/django_extensions/management/commands/runserver_plus.py", line 360, in <lambda>
  extra_files.extend(filter(lambda filename: filename.endswith('.mo'), gen_filenames()))
AttributeError: 'PosixPath' object has no attribute 'endswith'
```

But it seems to be just an issue with `pathlib`, you're expecting `str` and it gives you an instance of `PosixPath`.

https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/runserver_plus.py#L360

MRigal:
This is actually closed by #1320

trbs:
@MRigal thanks ! will close this :-)

PR #1320, opened 2019-02-22
Add workaround for extension checking on django22 with i18n

Qu4tro:
Following up https://github.com/django-extensions/django-extensions/issues/1295

Added a workaround on using `.endswith` on a Path (from pathlib). 

Since there are two possible datatypes for `filename` (str and PosixPath), I just wrap it around str(). 
In the case of str, we get the same string, in case of PosixPath, we get the path as a string.

Let me know, if there's anything I should add/modify.

Also, maybe too much ask, but can we get a release after? The current one on pypi, breaks due to this issue with the latest django version (https://pypi.org/project/django-extensions/)

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -357,7 +357,7 @@ class Command(BaseCommand):
             webbrowser.open(bind_url)
 
         if use_reloader and settings.USE_I18N:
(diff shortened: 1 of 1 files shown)


trbs:
no trouble at all 👍

trbs:
https://pypi.org/project/django-extensions/2.1.6/

Qu4tro:
Thank you!

PR #1325, opened 2019-02-26
Additional tests for find_template

kuter:
- added tests 
- removed template_loader workaround
- changed handle function

### DIFF
--- a/django_extensions/management/commands/find_template.py
+++ b/django_extensions/management/commands/find_template.py
@@ -19,4 +19,4 @@ class Command(LabelCommand):
         except TemplateDoesNotExist:
             sys.stderr.write("No template found\n")
         else:
-            print(template.name)
+            sys.stdout.write(self.style.SUCCESS((template.name)))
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/find_template.py
```diff
@@ -7,33 +7,16 @@
...
     help = "Finds the location of the given template by resolving its path"
     args = "[template_path]"
     label = 'template path'
 
     @signalcommand
     def handle_label(self, template_path, **options):
-        path = get_template_path(template_path)
-        if path is None:
+        try:
+            template = loader.get_template(template_path).template
```
trbs: Are we sure this does not break anything ?

django_extensions/management/commands/find_template.py
```diff
@@ -7,33 +7,16 @@
...
+        except TemplateDoesNotExist:
             sys.stderr.write("No template found\n")
-            sys.exit(1)
         else:
-            print(path)
+            print(template.name)
```
trbs: Maybe we can change this to use `self.stdout.write(self.style.SUCCESS(...))` at the same time ?

kuter:
@trbs take a look here, I've made changes based on your comments. If its ok .. then please merge that.

trbs:
Thanks !

Issue #1311, opened 2019-02-19
AutoSlugField method find_unique not overridable


SteinRobert:
Just had the case, in which I needed to override the `find_unique` method. 

I was creating a class from `AutoSlugField`:

```python
class MySlugField(AutoSlugField):
    def find_unique(model_instance, slug_field, generator):
        my_custom = model_instance.code
        ...
```

Unfortunately the method is never called, due to the current [implementation](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/fields/__init__.py#L196):

```python
        return super(AutoSlugField, self).find_unique(
            model_instance, slug_field, self.slug_generator(original_slug, start))
```

Is there a reason for this implementation? If not I'd be happy to submit a PR.

trbs:
Seems valuable to be able to override that method :-)
Happy to accept a PR

PR #1329, opened 2019-03-03
Replace `super().find_unique()` with `self.find_unique()`

SteinRobert:
Solves #1311

This way one is able to override `find_unique` as expected.


trbs:
Thanks for the PR !

Could you write a test so we can check the behavior works as expected and avoid future regressions ?

PR #1335, opened 2019-03-11
Additional test for notes command

kuter:


trbs:
Many thanks as always @kuter ! :-)

PR #1338, opened 2019-03-13
Additional tests for sync_s3 command

kuter:


kuter:
I've added boto requirement to tox.ini, console output says that it is installed .. 
`
py37-dj20 installed: atomicwrites==1.3.0,attrs==19.1.0,boto==2.49.0,certifi==2019.3.9,chardet==3.0.4,coverage==4.5.3,Django==2.0.13,django-extensions==2.1.6,django-pdb==0.6.2,factory-boy==2.11.1,Faker==1.0.4,idna==2.8,mock==2.0.0,more-itertools==6.0.0,pbr==5.1.3,pluggy==0.9.0,py==1.8.0,Pygments==2.3.1,pytest==4.3.1,pytest-cov==2.6.1,pytest-django==3.4.8,python-dateutil==2.8.0,pytz==2018.9,requests==2.21.0,shortuuid==0.5.0,six==1.12.0,text-unidecode==1.2,urllib3==1.24.1,Werkzeug==0.14.1
`
but within tests it raise ImportError. I've run these tests agains py37 and py27 on my local machine - all passed. @trbs could you check what might be wrong ? thanks.

django_extensions/management/commands/sync_s3.py
```diff
@@ -70,13 +70,17 @@
...
 
-# Make sure boto is available
-try:
-    import boto
-    import boto.exception
-    HAS_BOTO = True
-except ImportError:
-    HAS_BOTO = False
+
+def has_boto():
```
trbs: Is there a special reason for making this into a function ? personally I dislike any use of `global` when we can avoid it and it also makes the `import boto` non obvious.

kuter: yes, there is a reason .. I was unable to mock boto which was not installed.

trbs:
Thanks !

PR #1345, opened 2019-03-20
Tests for create_jobs management command

kuter:


trbs:
Thanks ! Your our coverage hero 🥇

PR #1360, opened 2019-04-25
runserver_plus: disable threading with --pdb, --ipdb, and --pm

blueyed:
This can otherwise lead to multiple pdb sessions in different threads,
especially when `pdb.set_trace()` is used additionally.


django_extensions/management/commands/runserver_plus.py
```diff
@@ -222,6 +222,13 @@ def postmortem(request, exc_type, exc_value, tb):
                 print("Exception occured: %s, %s" % (exc_type, exc_value), file=sys.stderr)
                 p.post_mortem(tb)
 
+        # Disable threading by default if debugger options are used.
+        if options["threaded"] is None:
+            if pdb_option or ipdb_option or pm:
+                options["threaded"] = False
```
blueyed: Thought about having a `print` there, but it should also get displayed once at most, which is not trivial it seems.

> print('Disabling threading with debugger, use --threaded to force.')

MRigal:
Don't know how to solve the `print` issue, but definitely 👍 for the issue!

blueyed:
Cool.  I don't think the print is really important, but maybe it should be mentioned with the option's help then - unfortunately there are two though.

django_extensions/management/commands/runserver_plus.py
```diff
@@ -72,7 +72,7 @@ def add_arguments(self, parser):
         parser.add_argument('--browser', action='store_true', dest='open_browser',
                             help='Tells Django to open a browser.')
         parser.add_argument('--nothreading', action='store_false', dest='threaded',
-                            help='Do not run in multithreaded mode.')
+                            help='Do not run in multithreaded mode.', default=None)
```
trbs: Please use the same style as above and put the `default` kwarg after `dest` instead of `help`.
Thanks

django_extensions/management/commands/runserver_plus.py
trbs: I like the idea of having the `print` there !

We should use:
```python
if self.show_startup_messages:
    print('Disabling threading with debugger, use --threaded to force.')
```

blueyed: I've tried that, but it did not work out like I've expected, but could retry / check it again.

trbs: > I've tried that, but it did not work out like I've expected, but could retry / check it again.

What was the issue with the `print` statement ?

blueyed: I think it was printed always (on reload) anyway - but might have been related to using this in another place / or it needs to be moved for this to work.

trbs: Would be nice if the output was there... (Hopefully it's not too difficult to add)
In any case it would be good if we update the `runserver_plus` docs as well and fix the small style issue below.

Let's get this merged :-)

django_extensions/management/commands/runserver_plus.py
blueyed: Amended.

trbs:
Thanks for the update.

Final request, can we please update the documentation of `runserverplus` to make mention of this ?

PR #1362, opened 2019-04-29
Added migrationstate management command

efficiosoft:
Hi,

This new management command can simplify migrating existing data to a new database greatly.

Your thoughts on including it in django-extensions are welcome. Tests will follow when you find it useful and think it can be included.

Best regards
Robert


docs/migrationstate.rst
```diff
@@ -0,0 +1,52 @@
...
+Caveats
+-------
+
+Consistency of Migrations
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+It should be clear, but I'm mentioning it anyway. To be able to reproduce
+a previously dumped migration state, all migration files up to that
+state still need to be present when restoring. Never delete or change
+migrations that have already been deployed.
```
trbs: This concluded `squashing` the migrations I assume ? If so would be good to mention that as well.

django_extensions/management/commands/migrationstate.py
```diff
@@ -0,0 +1,171 @@
+import json
+import sys
+
+from django.apps import apps
+from django.core.management import call_command
+from django.core.management.base import \
+    BaseCommand, CommandError, CommandParser
```
trbs: I would like to have this on a single line. Our `max-line-length` is quite high and honestly I find those `\`'s ugly :-)

docs/migrationstate.rst
```diff
@@ -0,0 +1,52 @@
...
+  $ ./manage.py migrationstate dump > migrationstate.json
+
+To dump the migration state of just a single app (`appname`)::
+
+  $ ./manage.py migrationstate appname > migrationstate.json
+
+To apply the migration state stored in `migrationstate.json`::
+
+  $ ./manage.py migrationstate load < migrationstate.json
+
```
trbs: If I understand correctly a common use case would be to do something like the following ?

```
 $ ./manage.py dumpscript auth > auth_dump.py
 $ ./manage.py migrationstate dump auth > auth_migrationstate.json
```

```
 $ ./manage.py migrationstate load < auth_migrationstate.json
 $ ./manage.py runscript auth_dump.py
```

Would be nice to also include such an example :-)

trbs:
Sounds useful :-)
Would be happy to include it.

I could imaging some additional features like a plain text format option and add an (optional) `-f` flag instead of in/output redirection.

These are all minor though...

What I would like to have are some tests for the command. This helps greatly with the maintainability of a command.

docs/migrationstate.rst
efficiosoft: It'll work as long as the migrations having the stored names are present, so squashing is ok as long as the original migrations are kept as well.

efficiosoft:
> Sounds useful :-)
> Would be happy to include it.

Great!

> I could imaging some additional features like a plain text format option and add an (optional) `-f` flag instead of in/output redirection.

Hmm, I don't think a plain text output is really needed since you can specify ``--indent=4`` to have it printed nicely, and there's still the native ``showmigrations`` command which - albeit serves a slightly different purpose - can output the same information pretty-printed.

Input/output file options are something I don't like very much in general for such simple commands... We have great shells these days and it would just add code to maintain. In addition, everybody, even if not familiar with the command, understands what ``<`` and ``>`` are going to do without having to consult the docs or call ``--help``. So I'd rather prefer not to add I/O file options, if that's okay to you.

> These are all minor though...

Glad to hear :)

> What I would like to have are some tests for the command. This helps greatly with the maintainability of a command.

Sure, just wanted to wait whether it's going to be accepted.

trbs:
No problem, thanks for the work on this !
When the tests land (and passes) I will merge this PR.

efficiosoft:
Just want to give a short feedback on this. I'm quite busy with other things at the moment. Please keep the PR open, I'll get back to it in a few weeks.

Best regards
Robert

efficiosoft:
Hi @trbs 

It turns out to be difficult to write tests for this command, because the tests have to roll migrations back and forth themselves, which isn't possible with current Django versions due to SQLite not supporting nested transactions.

What do you think, how should we proceed with this PR?

Best regards
Robert

trbs:
Would this be possible with `PostgreSQL` backend and if we only ran the tests on PG ?

efficiosoft:
Hmm, yes, that should work. But I wanted to avoid the overhead of setting up a local testenv just for that purpose... I'll see what I can do.

trbs:
Thank you !

I want to merge this PR but I feel tests would make it so much better :-)
Think the test for the `dump` are quite easy.

Maybe we can use the `--database` option to add another empty sqllite3 database to settings ? such that we can use that one to restore and test the various functions.

Issue #1374, opened 2019-05-27
Extend AutoSlugField ways of customize slugify_function 


chexex:
AutoSlugField already has the [ability](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/fields/__init__.py#L123) to customize default [slugify](https://docs.djangoproject.com/en/dev/ref/utils/#django.utils.text.slugify) function.

It is being said that you have to provide `slugify_function` as an argument to `AutoSlugField`. That was enough until I faced with a problem of changing a couple of [TitleSlugDescription](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/models.py#L45) models.

I think it would be great if we had an opportunity to provide `slugify_function` as an attribute within a model class.

Thanks

PR #1375, opened 2019-05-27
issue #1374 extend support for custom slugify function in AutoSlugFied

chexex:
The [issue](https://github.com/django-extensions/django-extensions/issues/1374)

Extended support for custom `slugify_function` in `AutoSlugField`. Added support for `slugify_function` attribute within a model class.

Added documentation
Added tests

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -99,6 +99,34 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
     overwrite
         If set to True, overwrites the slug on every save (default: False)
 
+    slugify_function
+        Defines the function which will be used to "slugify" a content
+        (default: :py:func:`~django.template.defaultfilters.slugify` )
+
+    It is possible to provide custom "slugify" function with
+    the ``slugify_function`` function in a model class.
+
+    ``slugify_function`` function in a model class takes priority over
+    ``slugify_function`` given as an argument to :py:class:`~AutoSlugField`.
+
+    Example
+
+    .. code-block:: python
+
+        # models.py
+
+        from django.db import models
+
+        from django_extensions.db.fields import AutoSlugField
+
+
+        class MyModel(models.Model):
+            def slugify_function(self, content):
+                return content.replace('_', '-').lower()
+
+            title = models.CharField(max_length=42)
+            slug = AutoSlugField(populate_from='title')
+
     Inspired by SmileyChris' Unique Slugify snippet:
     http://www.djangosnippets.org/snippets/690/
     """
@@ -143,9 +171,10 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
         value = re.sub('%s+' % re_sep, self.separator, value)
         return re.sub(r'^%s+|%s+$' % (re_sep, re_sep), '', value)
 
-    def slugify_func(self, content):
+    @staticmethod
+    def slugify_func(content, slugify_function):
         if content:
-            return self.slugify_function(content)
+            return slugify_function(content)
         return ''
 
     def slug_generator(self, original_slug, start):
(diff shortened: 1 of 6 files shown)


django_extensions/db/fields/__init__.py
```diff
@@ -143,9 +173,11 @@ def _slug_strip(self, value):
         value = re.sub('%s+' % re_sep, self.separator, value)
         return re.sub(r'^%s+|%s+$' % (re_sep, re_sep), '', value)
 
-    def slugify_func(self, content):
+    def slugify_func(self, content, model_instance):
         if content:
-            return self.slugify_function(content)
+            slugify_function = getattr(model_instance, 'slugify_function', None)
+            slugify_function = slugify_function or self.slugify_function
```
trbs: Why not use:
```python
slugify_function = getattr(model_instance, 'slugify_function', self.slugify_function)
```

Mental note: wondering about the performance impact of doing the `getattr` here instead of in `slug_generator`, though I don't really like that alternative as well since I then much rather also refactor the `lambda` out...

django_extensions/db/fields/__init__.py
```diff
@@ -99,6 +99,36 @@ class AutoSlugField(UniqueFieldMixin, SlugField):
...
+        from django.db import models
+
+        from django_extensions.db.fields import AutoSlugField
+
+
+        def my_slugify_function(self, content):
+            return content.replace('_', '-').lower()
+
+        class MyModel(models.Model):
+            slugify_function = my_slugify_function
```
trbs: What about using `class Meta` to specify the slugify function ?

```python
class MyModel(models.Model):
    title = models.CharField(max_length=42)
    slug = AutoSlugField(populate_from='title')

    class Meta:
        slugify_function = my_slugify_function
```

That feels a bit cleaner to me...

trbs:
I have no problem merging this but want to run a few idea by before doing so, see comments.

django_extensions/db/fields/__init__.py
chexex: Let me have a look on what we could do in the outer lamda function :) Thanks for the thoughts

django_extensions/db/fields/__init__.py
chexex: Agree

chexex: I did some research on **Meta** options.
Django has predefined [DEFAULT_NAMES](https://github.com/django/django/blob/master/django/db/models/options.py#L28) which are supposed to be the only attributes you can define within a **Meta** class.

<details><summary>The error if you try to define any custom attribute in <b>Meta</b></summary>
<code>
TypeError: 'class Meta' got invalid attribute(s): autoslugfield_slugify_function
</code>
</details>

The only way to bypass the problem is to redefine those **DEFAULT_NAMES** in a model's module where you are going to use the feature (which seems a bit cumbersome):
```py
from django.db import models
# ...
models.options.DEFAULT_NAMES = models.options.DEFAULT_NAMES + ('autoslugfield_slugify_function',)
```

If we decided to work with Meta, the overall code would look like this:
```py
"""
some models.py module
"""
from django.db import models

from django_extensions.db.models import TitleSlugDescriptionModel
from pytils.translit import slugify


# We have to write this line of code in each 
# module where we decided to use custom "slugify" feature
models.options.DEFAULT_NAMES = models.options.DEFAULT_NAMES + ('autoslugfield_slugify_function',)


class Author(TitleSlugDescriptionModel, models.Model):
    class Meta:
        autoslugfield_slugify_function = slugify
```

django_extensions/db/fields/__init__.py
chexex: - [x] Move getattr from ``slugify_func``
- [x] Added unit test for ``slugify_func``

django_extensions/db/fields/__init__.py
trbs: That definitely will not make it cleaner if we have to patch `models.options.DEFAULT_NAMES`.
Ah well.. it was a nice try :-) 


Think this would work as well right ?
```python
class MyModel(models.Model):
    title = models.CharField(max_length=42)
    slug = AutoSlugField(populate_from='title')

    def slugify_function(self, content):
        return content.replace('_', '-').lower()
```
or
```python
class MyModel(models.Model):
    title = models.CharField(max_length=42)
    slug = AutoSlugField(populate_from='title')

    def slugify_function(self, content):
        return slugify(content)
```

Which would probably be my preferred way of writing it now that `Meta` does not work :-)
Also makes it nice for creating mixin classes etc.

chexex: > Think this would work as well right ?

Correct.
I altered docs and examples.
Indeed, function looks way more attractive :)

trbs:
Awesome ! Thanks !

PR #1388, opened 2019-06-26
Add arrow_shape argument in graph_models

Offpics:
This pull requests adds new optional argument '--arrow-shape' in graph_models. It allows to change arrow shape of a relation from 'dot' to other [alternatives](https://www.graphviz.org/doc/info/arrows.html).

I think it would be a great addition. I don't like the dot arrow and I had to change it manually in text editor after generating .dot file with graph_models.

### DIFF
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -168,10 +168,10 @@ class Command(BaseCommand):
             },
             '--arrow-shape': {
                 'action': 'store',
+                'default': 'dot',
                 'dest': 'arrow_shape',
-                'help': 'Arrow shape to use for relations. \
-                    Default is dot. Available shapes: box, crow, \
-                    curve, icurve, diamond, dot, inv, none, normal, tee, vee.',
+                'choices': ['box', 'crow', 'curve', 'icurve', 'diamond', 'dot', 'inv', 'none', 'normal', 'tee', 'vee'],
+                'help': 'Arrow shape to use for relations. Default is dot. Available shapes: box, crow, curve, icurve, diamond, dot, inv, none, normal, tee, vee.',
             }
         }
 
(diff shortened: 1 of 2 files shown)


trbs:
I love f-strings but they require a little more modern Python version then we support at the moment. 
Until Django 1.11 LTS expires in April 2020 we cannot use them :(

django_extensions/management/commands/graph_models.py
```diff
@@ -165,6 +165,13 @@ def __init__(self, *args, **kwargs):
                 'default': False,
                 'dest': 'hide_edge_labels',
                 'help': 'Do not showrelations labels in the graph.',
+            },
+            '--arrow-shape': {
+                'action': 'store',
+                'dest': 'arrow_shape',
+                'help': 'Arrow shape to use for relations. \
+                    Default is dot. Available shapes: box, crow, \
+                    curve, icurve, diamond, dot, inv, none, normal, tee, vee.',
```
trbs: Lets use `argparse` `choices` option here.

We can copy the `shapes` list in` modelviz.py` as a constance here.

That creates couple of improvements:

(1) we don't need the `\` line continuations anymore (which I always find ugly to look at) since it will fit on one line. 

(2) The command will check that only valid choices are allowed. 

(3) We can remove the `parse_arrow_shape` error check since now only valid options can be passed down anyways.

What do you think ?

Offpics: I wasn't aware of the `argparse` `choices` and it seems to be superior choice here. I will change it.

Long comments in one line are really hard to read for me, but sure I will change it to one line.

trbs: Variation could be something like:

```
        arrow_shapes = ['box', 'crow', 'curve', 'icurve', 'diamond', 'dot', 'inv', 'none', 'normal', 'tee', 'vee']
...
                'choices': arrow_shapes,
                'help': 'Arrow shape to use for relations. Default is dot. Available shapes: %s' % ', '.join(arrow_shapes),
```

Then the shapes are specified once and the 'help' line is a bit sorter since it does not need to list them all again.

There also some other ways to do multi line text without using `\`.

But anyways this is more then fine enough for me :-)

Issue #1394, opened 2019-07-17
MySQL 8 returns upper case column names results in KeyError in dictionary


potulabe:
In function [sql_to_dict](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/sqldiff.py#L257) SELECT result which was returned from sql-server is transformed to dictionary. Then some particular keys are supposed to be in this dict, dependent on SELECT, for example, ["column_name"](https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/commands/sqldiff.py#L768) . But MySQL 8 ( Ver 8.0.15-6 for debian-linux-gnu on x86_64 (Percona Server (GPL), Release '6', Revision '63abd08')) returns column names in upper case, so instead "column_name" there will be "COLUMN_NAME".

Possible solution is to add .lower() to name[0] in sql_to_dict()

trbs:
If you want to make a PR for this I think the easiest way is to override the `sql_to_diff` method in the `MySQLDiff` class to support this. 

I am not familiar with MySQL would we need to do this only for MySQL 8 or also for other mysql versions or variants (MariaDB et al) ?

PR #1399, opened 2019-07-25
Ticket-1394- Added Mysql format_field_names testcase

LanghaniKuldeep:
connects https://github.com/django-extensions/django-extensions/issues/1394
Test Case for format_field_name for MySqlDiff Class.

### DIFF
--- a/tests/test_sqldiff.py
+++ b/tests/test_sqldiff.py
@@ -6,7 +6,7 @@ from django.apps import apps
 from django.test import TestCase
 
 # from django.core.management import call_command
-from django_extensions.management.commands.sqldiff import SqliteSQLDiff, Command
+from django_extensions.management.commands.sqldiff import SqliteSQLDiff, Command, MySQLDiff
 
 
 class SqlDiffTests(TestCase):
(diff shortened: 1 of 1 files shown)


nasirhjafri:
@LanghaniKuldeep Thanks for test case.

trbs:
Thanks for the PR.

Two questions:

1) This test can run without the need for `@pytest.mark.skipif(settings.DATABASES['default']['ENGINE'] != 'django.db.backends.mysql',...` since it's straight up testing the `format_field_names` method not doing any database calls. I added a simple patch to master to facilitate this.

2) Can we also write a test that tests `sql_to_dict` on `sqlite`, `postgresql` and `mysql`. We have test models that we can `SELECT` from but I assume something like `.sql_to_dict("""select 1 as "foo", 1 + 1 as "BAR";""",[])` will also work.

LanghaniKuldeep:
@trbs Added test case according to given conditions.

tests/test_sqldiff.py
```diff
@@ -6,23 +6,27 @@
...
+
     def _include_proxy_models_testing(self, should_include_proxy_models):  # type: (bool) -> ()
-        parser = Command().create_parser("test", "sqldiff")
-        args = ["-a"]
         if should_include_proxy_models:
-            args.append("--include-proxy-models")
-        options = parser.parse_args(args=args)
-        tmp_out = six.StringIO()
-        tmp_err = six.StringIO()
+            self.args.append("--include-proxy-models")
```
trbs: This should be `args` not `self.args` ?
Now we keep appending to `self.args` which each test executed in the class.

LanghaniKuldeep: @trbs I assumed that SetUp is executed for each tests method case separately.

LanghaniKuldeep: I only need that in this case only therefore only appended here.

tests/test_sqldiff.py
```diff
@@ -36,6 +40,42 @@ def test_sql_diff_without_proxy_models(self):
...
+        instance = MySQLDiff(
+            apps.get_models(include_auto_created=True),
+            vars(self.options),
+            stdout=self.tmp_out,
+            stderr=self.tmp_err,
+        )
+        expected_field_name = ['name', 'email', 'address']
+        self.assertEquals(instance.format_field_names(['Name', 'EMAIL', 'aDDress']), expected_field_name)
+
+    def test_sql_to_dict(self):
```
trbs: I would like to split these up in 3 tests for each database and have the `@pytest.skipif(...)` decorator on them to test only when that database engine is used.

This because even though the test right now works, it does not really test what you would expect. In most of our test runs no matter which Diff engine is used it will always run on `sqlite` since that it the database engine for most test runs.

Also I would love to test the failure case as well. That way we can actually see that this PR is useful and it's easier to test regressions or find to find out that database engine change their behaviors.

So I think this test should be split in 3, one for each database server and it should monkeypatch and `assertEquals` the failure case in MySQL as well.

trbs:
Thanks for all the hard work on the tests !

I added a few comments to make the test even more awesome :-)

tests/test_sqldiff.py
trbs: Your absolutely right. It's called before every test... makes sense to otherwise `tmp_out` and `tmp_err` would have been problematic as well.

(Guess in the end I would probably prefer to rewrite these into `pytest` style tests but that's for another day :-) )

trbs:
Thanks !

PR #1400, opened 2019-07-27
Updated GoDjango video url

avenet:
GoDjango site does not exist anymore, so the URL for the screencast was not working.


trbs:
Thanks !

Issue #1371, opened 2019-05-19
reset_db in 2.1.7 breaks PostGIS


jensenbox:
```
DATABASES = {
    'default': {
        'ENGINE': 'django.contrib.gis.db.backends.postgis',
        "NAME": "dev",
        "USER": "postgres",
        "HOST": "db",
        "PORT": 5432,
    }
}
```

`manage.py reset_db --close-sessions --noinput`

`CommandError: Unknown database engine django.contrib.gis.db.backends.postgis`

I have a thought it has something to do with the new functionality of being able to change the driver

Pinning to 2.1.6 does not exhibit the problem.

jensenbox:
I tested by adding it to the setting to override and it worked.

```
DJANGO_EXTENSIONS_RESET_DB_POSTGRESQL_ENGINES = (
            'django.db.backends.postgresql',
            'django.db.backends.postgresql_psycopg2',
            'django.db.backends.postgis',
            'psqlextra.backend',
            'django.contrib.gis.db.backends.postgis'
        )
```

trbs:
I think `django.db.backends.postgis` is a typo...

I've committed `43a7da4c44f58fb85db2edab9bb91bd54b01e23e` could you please check with master ?

jdzero:
https://github.com/django-extensions/django-extensions/blob/b2b60725f3bf0330720f0eccd6fd844e6e58c52e/django_extensions/settings.py#L13

I think the intent was for this to be innocuous -- I believe the gis driver for mysql should be in this list.

PR #1402, opened 2019-08-05
Adds missing GIS engine for MySql to defaults.

jdzero:
To address:

https://github.com/django-extensions/django-extensions/issues/1371


trbs:
Thanks !

PR #1404, opened 2019-08-08
sqldiff - mysql code can duplicate AUTO_INCREMENT and UNSIGNED 

### LINKED ISSUE #1403: MySQL sqldiff duplicates AUTO_INCREMENT and UNSIGNED declarations creating false positives
e.g.

`App auth Model Permission field 'id' parameters differ: db='integer AUTO_INCREMENT AUTO_INCREMENT', model='integer AUTO_INCREMENT'`

xurble:
Fixes #1403 

Only tested on MySQL but looks quite safe??

### DIFF
--- a/django_extensions/management/commands/sqldiff.py
+++ b/django_extensions/management/commands/sqldiff.py
@@ -354,7 +354,7 @@ class SQLDiff(object):
         tablespace = field.db_tablespace
         if not tablespace:
             tablespace = "public"
-        if (tablespace, table_name, field.column) in self.unsigned:
+        if (tablespace, table_name, field.column) in self.unsigned and self.unsigned_suffix not in field_db_type:
             field_db_type = '%s %s' % (field_db_type, self.unsigned_suffix)
 
         return field_db_type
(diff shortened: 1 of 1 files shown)


django_extensions/management/commands/sqldiff.py
```diff
@@ -354,7 +354,7 @@ def get_field_db_type(self, description, field=None, table_name=None):
         tablespace = field.db_tablespace
         if not tablespace:
             tablespace = "public"
-        if (tablespace, table_name, field.column) in self.unsigned:
+        if (tablespace, table_name, field.column) in self.unsigned and self.unsigned_suffix not in field_db_type:
```
trbs: Have not tested it but should we use `... and (self.unsigned_suffix is not None and self.unsigned_suffix not in field_db_type):` here ?

To avoid `TypeError: 'in <string>' requires string as left operand, not NoneType`

xurble: `unsigned_suffix` is a class variable and essentially a constant so it should be safe, but if you'd rather have belt and braces that's fine too.

trbs: I was looking that it's only set in `MySQLDiff` and otherwise set to `None` resulting in `TypeError`.
However if it's set to `None` then the `self.unsigned` set would also be empty :-)

Issue #1392, opened 2019-07-05
UserPassesTestMixin for currently logged in user


tdsprogramming:
I have a mixin I often need to use that is built off `django.contrib.auth.mixins.UserPassesTestMixin` that limits permission of the currently logged in user to only be able to change models that they own. It looks something like this:
```python
class UserPermissionMixin(UserPassesTestMixin):
    def test_func(self):
        current_user = self.request.user
        if current_user == self.model.user:
            return True
        return False
```
Would this be something that I could generalize and submit a pull request for django-extensions?

trbs:
I don't think we have any permissions yet in `django-extensions` :-)

Thinking about it, from personal experience I have a similar mixins which are copied in quite a number of projects. Others probably do the same making it a good candidate to include IMO.

What I would like to see if the ability to change the model attribute. Something like:
```
class UserPermissionMixin(AccessMixin):
    user_permission_model_attribute = "user"

    def get_user_permission_model_attribute(self):
        return self.user_permission_model_attribute

    def test_func(self):
        model_attr = self.get_user_permission_model_attribute()
        current_user = self.request.user
        model_user = getattr(self.model, model_attr)
        current_user = self.request.user

        if current_user == model_user:
            return True

        return False
```

This would allow for more flexibility. eg: you could have a class with an `owner` attribute.
Maybe also add the ability for `user_permission_model_attribute` to be a relationship, such as in the case of model with something like an `attendees` attribute.

Also a similar class for `groups` would make sense :-)

Anyways, I would love to accept a PR for this. Specially if we can make the class generic and easy to use enough that it would appeal to a broad number of users.

Now the only downside to a PR right now is that it would also require some of the boilerplate for things like docs, tests (we should aim for 100% coverage for permission mixins), etc to be written since this is the first permissions extension.

tdsprogramming:
Great! Where would be an appropriate place for this? Since the structure is similar to Django, should it be from `django_extensions.contrib.auth.mixins import UserPermissionMixin`?

trbs:
Think we can skip the `contrib` part and go with `django_extensions.auth.mixins` :-)

tdsprogramming:
I'm writing up the docs, but I'm not comfortable writing tests for django extensions. Can I submit the pull request with docs and source code, and then get an idea for the tests after?

tdsprogramming:
[...]

PR #1405, opened 2019-08-08
Permissions

tdsprogramming:
Added a `UserPermissionMixin` that limits access to a model to the currently logged in User. Also added docs and, but no tests.


django_extensions/auth/mixins.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.auth.mixins import AccessMixin
+
```
trbs: The build is failing on a very minor whitespace thing here :-) 
`pep8` says top level functions and classes should be separated with two blank lines instead of one.

trbs:
The PR looks very good, thanks for your work on this !

Maybe we should call it `ModelUserFieldPermissionMixin` or `ModelOwnerPermissionMixin` ? 
Which is kinda long but does more explicitly describes what the Mixin is doing :-) (In the sense that `UserPermissionMixin` might be confused with other types/ways of User permissions.)

What do you think about tests ?  Maybe we can look at Django's `tests/auth_tests/test_mixins.py` and implement basic tests here as well.

tdsprogramming:
Good idea for the tests. I'll check out Django's docs and fix that on the next go around. I like `ModelUserFieldPermissionMixin` more because it adds more of a generic name.

tdsprogramming:
I need a model with a user field for testing. Is there one in `tests/testapp/models.py`? Can I write one there or should I make a model just in the test file?

trbs:
Either is fine for me :-)

If the model can be used more broadly and for other tests (now or in the future) I would prefer `tests/testapp/models.py`, otherwise if there very specific to these test use the test file :-)

anudeepsamaiya:
adding reference to the issue #1392

tdsprogramming:
About to push with tests. I essentially used the same things from Django's tests, per our previous discussion: https://github.com/django/django/blob/master/tests/auth_tests/test_mixins.py#L195. Not sure if these tests will be effective enough

tdsprogramming:
Please ignore all previous commits; I realized I had a major issue that I am working through. I will push once that is resolved.

tdsprogramming:
The most recent push should be working.

trbs:
> The most recent push should be working.

sorry seems there still at least some style issues...

tdsprogramming:
Where can I find the style issues?

trbs:
https://travis-ci.org/django-extensions/django-extensions/jobs/605077653?utm_medium=notification&utm_source=github_status

trbs:
Thanks !

tdsprogramming:
No problem. I realized I think I forgot to rebase a couple times. I don't think anything was messed up though, correct?

PR #1406, opened 2019-08-13
Add a templatetag for breakpoint() debugging

danihodovic:
PEP 553 introduced the breakpoint() builtin which is short for

```
import pdb; pdb.set_trace()
```

it either uses the default debugger or whatever is set in $PYTHONBREAKPOINT
e.g to use pudb
```
export PYTHONBREAKPOINT=pudb.set_trace
```

Usage should be `{% debugger %}`


danihodovic:
I've looked into adding tests, but I'm unsure how to mock a builtin method (breakpoint())

PR #1410, opened 2019-08-28
Graph-models now permit the combination of includes and excludes

### LINKED ISSUE #1409: Cannot combine --exclude and --include with graph_models
## What I'm trying to achieve
I would like to be able to include some models through given wildcards. But yet be able to include some of them against given wildcards.

## Example
Let's say I want to only select some given models and all the models containing the word `Attribute` but yet, exclude any containing the word `Translation`. I would do the following:
```
./manage.py graph_models product \
    -I '*Attribute*,Product,ProductVariant,ProductType' \
    -X '*Translation*' \
    --pygraphviz -g -o attributes_class_tree.png
```

## What I expected to happen
1. Have the models matched against the exclude list
1. If the model didn't match, match against the include list

## What happened instead
Nothing was selected, thus nothing was rendered.

## Suggested solution
```diff
Index: django_extensions/management/modelviz.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- django_extensions/management/modelviz.py	(revision 52cbf38144109a705f6932dd996540b66f014494)
+++ django_extensions/management/modelviz.py	(date 1567019406336)
@@ -381,18 +381,18 @@
         Decide whether to use a model, based on the model name and the lists of
         models to exclude and include.
         """
+        # Check against include list.
+        if self.include_models:
+            for model_pattern in self.include_models:
+                model_pattern = '^%s$' % model_pattern.replace('*', '.*')
+                if re.search(model_pattern, model_name):
+                    return True
         # Check against exclude list.
         if self.exclude_models:
             for model_pattern in self.exclude_models:
                 model_pattern = '^%s$' % model_pattern.replace('*', '.*')
                 if re.search(model_pattern, model_name):
                     return False
-        # Check against exclude list.
-        elif self.include_models:
-            for model_pattern in self.include_models:
-                model_pattern = '^%s$' % model_pattern.replace('*', '.*')
-                if re.search(model_pattern, model_name):
-                    return True
         # Return `True` if `include_models` is falsey, otherwise return `False`.
         return not self.include_models
```

<p align='center'>(snippet from <a href='https://github.com/django-extensions/django-extensions/blob/d6d07584004f2a1fa0734edfc144fcf37f33ac4b/django_extensions/management/modelviz.py#L379-L397'>management/modelviz.py</a>)</p>

---

I will open a pull request later on today, with the [adequate tests](https://github.com/django-extensions/django-extensions/blob/7ac4792c72338c95d4e2865a786504b28c157383/tests/test_management_command.py#L127-L222). Let me know about what you think.

NyanKiyoshi:
This change allows users to provide both --include (-I) and --exclude (-X) to filter given wildcard patterns, while being still able to only include given model names or given wildcards to match.

For example, one might want to exclude all models, using --include, then filter out some unwanted models through --exclude, e.g. M2M serving as translation service.

This would result to the following usage:
```
./manage.py -I 'Attribute*' -X '*Translation' <...>
```

Which basically tells graph-models to exclude all models, and only keep the ones having their name starting by "Attribute" and remove any of those that ends by the word "Translation".

---

Closes #1409.

### DIFF
--- a/django_extensions/management/modelviz.py
+++ b/django_extensions/management/modelviz.py
@@ -381,18 +381,18 @@ class ModelGraph(object):
         Decide whether to use a model, based on the model name and the lists of
         models to exclude and include.
         """
+        # Check against include list.
+        if self.include_models:
+            for model_pattern in self.include_models:
+                model_pattern = '^%s$' % model_pattern.replace('*', '.*')
+                if re.search(model_pattern, model_name):
+                    return True
         # Check against exclude list.
         if self.exclude_models:
             for model_pattern in self.exclude_models:
                 model_pattern = '^%s$' % model_pattern.replace('*', '.*')
                 if re.search(model_pattern, model_name):
                     return False
-        # Check against exclude list.
-        elif self.include_models:
-            for model_pattern in self.include_models:
-                model_pattern = '^%s$' % model_pattern.replace('*', '.*')
-                if re.search(model_pattern, model_name):
-                    return True
         # Return `True` if `include_models` is falsey, otherwise return `False`.
         return not self.include_models
 
--- a/tests/test_management_command.py
+++ b/tests/test_management_command.py
@@ -138,6 +138,7 @@ class GraphModelsTests(TestCase):
             '*WildcardPrefixExclude',
             'WildcardSuffixExclude*',
             '*WildcardBothExclude*',
+            '*Include',
         ]
         # Any model name should be used if neither include or exclude
         # are defined.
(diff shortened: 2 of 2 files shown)


trbs:
Thanks !

If you can would you be willing to make a PR against the documentation to describe the exclude overwrites includes behavior ?

NyanKiyoshi:
@trbs thanks for the merge :+1: I will do that as soon as I can.

Issue #1312, opened 2019-02-19
print_user_for_session doesn't use auth backend from session


schwobaseggl:
Is there a concrete reason why the following is used:

```
uid = data.get('_auth_user_id', None)

User = get_user_model()
try:
    user = User.objects.get(pk=uid)
except User.DoesNotExist:
    # ....
```

wouldn't it be better to actually use the backend that logged the user in. Note that the logged-in user does not need to be an instance of `get_user_model()` because custom auth backends allow for different models. Sth along the following lines should work just as fine and cover these cases:

```
from django.utils.module_loading import import_string

uid = data.get('_auth_user_id', None)
backend_path = data.get('_auth_user_backend', None)

try:
    backend = import_string(backend_path)()
    user = backend.get_user(user_id=uid)
except (ObjectDoesNotExist, ImportError):
    # ....
```

The model itself would then be valuable information for the output as well:
```
opts = user._meta
print('{}.{}'.format(opts.app_label, opts.model_name))
```

If that seems reasonable, I will be glad to provide a pull request.

trbs:
Makes sense, any chance I can convince you to create a PR for this ? ;-)

PravinKamble123:
Hey, I’d love to work on this issue. Could you assign it to me?

PravinKamble123:
Thanks for assigning this! 
I had a look and noticed there’s already a merged PR (#1419) that addresses it.
Happy to help if there’s anything else needed!

PR #1419, opened 2019-09-11
Feat: use session backend on print user for session command

karranb:
#1312

This PR adapts the command 'print_user_for_session' to use the session's backend model.

### DIFF
--- a/django_extensions/management/commands/print_user_for_session.py
+++ b/django_extensions/management/commands/print_user_for_session.py
@@ -4,7 +4,6 @@ from __future__ import print_function, unicode_literals
 import importlib
 
 from django.conf import settings
-from django.contrib.auth import get_user_model
 from django.contrib.sessions.backends.base import VALID_KEY_CHARS
 from django.core.management.base import BaseCommand, CommandError
 from django.utils.module_loading import import_string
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/print_user_for_session.py
```diff
@@ -39,19 +39,18 @@ def handle(self, *args, **options):
...
             print('No user associated with session')
             return
 
         print(u"User id: %s" % uid)
 
-        User = get_user_model()
-        try:
-            user = User.objects.get(pk=uid)
-        except User.DoesNotExist:
+        backend = import_string(backend_path)()
```
trbs: Let's use `from django.contrib.auth import load_backend`

django_extensions/management/commands/print_user_for_session.py
```diff
@@ -39,19 +39,18 @@ def handle(self, *args, **options):
 
         print('Session to Expire: %s' % session.get_expiry_date())
         print('Raw Data: %s' % data)
-
         uid = data.get('_auth_user_id', None)
+        backend_path = data.get('_auth_user_backend', None)
```
trbs: Can you change this to `from django.contrib.auth import BACKEND_SESSION_KEY, SESSION_KEY` ?

(I suggest fixing `_auth_user_id` at the same time :-) )

We also need an `if backend_path is None:` in case for whatever reason `_auth_user_backend` does not exist in the session and give some feedback to the user.

trbs:
Thanks ! I left few little notes how we can (hopefully) make it even better :-)

karranb:
Hi @trbs, awesome improvements. I implemented your suggestions and added a simple test to it.

trbs:
awesome ! Thanks for the PR @karranb !

PR #1420, opened 2019-09-11
Updated docs to mention the possibility of combining include and exclude

NyanKiyoshi:
Ref: https://github.com/django-extensions/django-extensions/pull/1410


trbs:
Awesome thanks !

PR #1430, opened 2019-10-15
Support unique_together with RandomCharField

swt2c:
AutoSlugField already supports unique_together through find_unique() in UniqueFieldMixin.  Add support for unique_together in RandomCharField as well by checking for the attribute name in unique_together before short-circuiting find_unique() there.


trbs:
Thanks !!

Issue #1425, opened 2019-09-28
Add Indexes for TimeStampedModel on created and modified fields.


anudeepsamaiya:
By default indexes should be added inside the Meta of abstract class.

vstoykov:
Can you provide a PR?

trbs:
Also in the future please provide a little bit of reasoning behind the request or PR :-)

I assume this is because that way if one overrides the class you can easily change the indexes to your liking.

Happy to accept a PR for this.

anudeepsamaiya:
Yes, I have almost every model in my project inherit from `TimeStampedModel`. So it would be nice to have indexes in `AbstractModel` itself. 

I was waiting to raise a PR.

trbs:
@anudeepsamaiya if you submit a PR I will try to much it as fast as I can and make a new release for you. (Next month I will be traveling a lot and probably not get to it.)

PawelRoman:
TimeStampedModel seems to have a design flaw: it adds two non-indexed fields (created and modified) and at the same time it defines a default ordering on those fields. That results in a full table scan each time a query is run on any model that inherits from TimeStampedModel.

Since explicit is better than implicit I would simply remove the default ordering from the TimeStampedModel.

trbs:
> TimeStampedModel seems to have a design flaw: it adds two non-indexed fields (created and modified) and at the same time it defines a default ordering on those fields. That results in a full table scan each time a query is run on any model that inherits from TimeStampedModel.
> 
> Since explicit is better than implicit I would simply remove the default ordering from the TimeStampedModel.

Not sure if it's a design flaw per see, but at the least it can be surprising to people implementing the abstract class. 

I agree... seems that explicitly defining ordering on the model using the `TimeStampedModel` is a better approach. 

Thinking that indexes might be similar and one can always make a small base abstract class of their own if it's used a lot; eg: `IndexedOrderedTimeStampedAbstractModel(TimeStampedModel)` :-)

For now I have removed the default ordering and will close this issue for now.

If there is enough need in the world to add a subclass with index support I'm happy to add that put please first make you get some decent number of developers / users to support your change before submitting a PR.

Thanks to everybody contributing to this issue !

PR #1433, opened 2019-10-23
add indexes to timestamped models 

anudeepsamaiya:
Issue Ref: django-extensions/django-extensions#1425

Added indexes in the `Meta` of the `TimeStampedModel`.

Usage:
```
class Cheese(TimeStampedModel):
    name = models.CharField(max_length=50, null=True, blank=True)
```
or
```
class Sale(TimeStampedModel):
    quantity = models.IntegerField()

    class Meta(TimeStampedModel.Meta):
        db_table = "sale"
```

or 

```
class Sale(TimeStampedModel):
    quantity = models.IntegerField()

    class Meta(TimeStampedModel.Meta):
        db_table = "sale"
         indexes = TimeStampedModel.Meta.indexes + [models.Index(fields=["quantity"])]
```

### DIFF
--- a/django_extensions/db/models.py
+++ b/django_extensions/db/models.py
@@ -25,6 +25,7 @@ class TimeStampedModel(models.Model):
     class Meta:
         get_latest_by = 'modified'
         ordering = ('-modified', '-created',)
+        indexes = [models.Index(fields=["created"]), models.Index(fields=["modified"])]
         abstract = True
 
 
--- a/django_extensions/db/models.py
+++ b/django_extensions/db/models.py
@@ -25,7 +25,21 @@ class TimeStampedModel(models.Model):
     class Meta:
         get_latest_by = 'modified'
         ordering = ('-modified', '-created',)
-        indexes = [models.Index(fields=["created"]), models.Index(fields=["modified"])]
+        abstract = True
+
+
+class IndexedTimeStampedModel(TimeStampedModel):
+    """
+    IndexedTimeStampedModel
+
+    An abstract base class model that provides self-managed "created" and
+    "modified" fields with db indices on them.
+    """
+    class Meta(TimeStampedModel.Meta):
+        indexes = [
+            models.Index(fields=["-created", "-modified"]),
+            models.Index(fields=["-created"]), models.Index(fields=["-modified"])
+        ]
         abstract = True
 
 
@@ -40,6 +54,20 @@ class TitleDescriptionModel(models.Model):
     description = models.TextField(_('description'), blank=True, null=True)
 
     class Meta:
+        abstract = True
+
+
+class IndexedTitleDescriptionModel(TitleDescriptionModel):
+    """
+    IndexedTitleDescriptionModel
+
+    An abstract base class model that provides title and description fields
+    with an index on title field.
+    """
+    class Meta(TitleDescriptionModel.Meta):
+        indexes = [
+            models.Index(fields=["title"]),
+        ]
         abstract = True
 
 
@@ -61,6 +89,22 @@ class TitleSlugDescriptionModel(TitleDescriptionModel):
     slug = AutoSlugField(_('slug'), populate_from='title')
 
     class Meta:
+        abstract = True
+
+
+class IndexedTitleSlugDescriptionModel(TitleSlugDescriptionModel):
+    """
+    IndexedTitleSlugDescriptionModel
+
+    An abstract base class model that provides title and description fields
+    and a self-managed "slug" field that populates from the title with indexes
+    on "title" and "slug" fields.
+    """
+    class Meta(TitleSlugDescriptionModel.Meta):
+        indexes = [
+            models.Index(fields=["title", "slug"]),
+            models.Index(fields=["title"]), models.Index(fields=["slug"])
+        ]
         abstract = True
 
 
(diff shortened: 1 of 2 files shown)


trbs:
Could you please update `docs/model_extensions.rst` and put the usage example in the description as an example under `*TimeStampedModel*`.

anudeepsamaiya:
> Could you please update `docs/model_extensions.rst` and put the usage example in the description as an example under `*TimeStampedModel*`.

I have added the docs. Thanks.

django_extensions/db/models.py
```diff
@@ -25,6 +25,7 @@ def save(self, **kwargs):
     class Meta:
         get_latest_by = 'modified'
         ordering = ('-modified', '-created',)
+        indexes = [models.Index(fields=["created"]), models.Index(fields=["modified"])]
```
vstoykov: Currently these indexes are like adding `db_index=True` to each of the fields (`created` and `modified`). There is no need for `Meta.indexes` in this case. 

If `Meta.indexes` should be defined I thing that is better to add index which is for both fields in descending order (to match the `Meta.ordering`). This way listing of the objects should be faster. Also individual indexes can also be in descending order (at least for `modified`) for `latest()` to use it.

docs/model_extensions.rst
```diff
@@ -37,7 +37,41 @@ Both of the fields are customly defined in Django Extensions as
...
+To override the ``Meta`` class in the concrete class with new options, then
+``Meta`` of the concrete class should inherit from ``TimeStampedModel.Meta``.
+
+::
+  class Sale(TimeStampedModel):
+      quantity = models.IntegerField()
+
+      class Meta(TimeStampedModel.Meta):
+          db_table = "sale"
+
```
vstoykov: When explicitly extending a `Meta` class from abstract model and you do not want abstract class then `abstract = False` should be set explicitly if I'm not mistaken.

anudeepsamaiya: No Django takes care of that, it sets the flag to `false`, but if you want the child model to be abstract as well then you have to set it to `true` explicitly.

django_extensions/db/models.py
anudeepsamaiya: I agree with your points. So do you think it would be better if the modified the current version like this:
```
indexes = [ 
    models.Index(fields=["-created", "-modified"]),
    models.Index(fields=["-created"]), models.Index(fields=["-modified"])
]
```

anudeepsamaiya:
> 
> Why not do a little differently... Add an `IndexedTimeStampedModel` (name is just a strawman) which is something like:
> 
> ```python
> class IndexedTimeStampedModel(TimeStampedModel):
>     class Meta(TimeStampedModel.Meta):
>         indexes = [ 
>             models.Index(fields=["-modified"])
>             models.Index(fields=["-created"]),
>         ]
> ```
>
I was thinking of implementing this way, but then most people will have to modify their code if they want to use indexes. On the other hand if we go `TimeStampedModel` any new user will have indexes right from start and others will have database migrations. The latter one will surely introduce inconveniences but in favor of performance improvements.

Although, I am not in favor of going with latter options, we can totally avoid it if we by adding the document for `IndexedTimeStampModel` and properly explaining its pros and cons. I also think we can also have indexed versions for other models. 

> 
> (or `models.Index(fields=["-modified", "-created"])` we have to decide on one multi column index or two separate ones. Do all supported django database offer multi column indexes ?)
>
Yes `MySQL`, `MariaDB`, `Postgres`, `Oracle`, `SQLServer` all popular dbs support multi column indexes. 

> 
> This way we do not suddenly change the behavior of `TimeStampedModel` and create a schema migration after an upgrade of `django-extensions` for everybody.

trbs:
Cool lets do it that way and add index'd versions with some documentation.

This would also be a very nice example of how easy it can be to customize these things :-)

docs/model_extensions.rst
vstoykov: Mind blowing. I'm almost sure that had e a case where I had to put `abstract = False` explicitly on a `Meta` which was extending another `Meta`. Probably there was some other reason for that.

YPCrumble:
FWIW as a django-extensions user I think adding the index to my existing models, vs. having to make complex migrations to migrate my `TimeStampedModel` columns to `IndexedTimeStampedModel` columns, is the right way to go...per what @anudeepsamaiya said here:

> On the other hand if we go with adding indexes to TimeStampedModel, then any new user will have indexes right from start and others will have database migrations.

This would be _much_ easier for me than having to switch my existing `TimeStampedModel`'s to `IndexedTimeStampedModels`. 

@trbs I would think that for almost all use cases, having an index on `created` and `modified` fields that are used so often in sorting would be expected by default. See for instance [this SO question/answers](https://stackoverflow.com/questions/15425230/is-it-a-good-idea-to-index-datetime-field-in-mysql).

For large projects this would (I believe) significantly speed up queries, and for small projects this would make a negligible difference from the current implementation.

trbs:
Unless you have `TimeStampedModel` on very large tables where you don't need the index and it will slow down the database, bloat the size (because of the additional indexes) and potentially mess a little with caching since you now evicting potentially hot pages :-)

Anyways so many potential cases and we really don't know what people are using it for.

I can be convinced to set the indexes by default, specially since I think having a complex migration path for people (I assume the majority) that do want to have the indexes by default.

So what I suggest to do is if there is consensus to make the indexes the default. Then we do so via:

1. The `class Meta` `indexes` property
2. Update the PR to fix the style of those as indicated in my comments/reviews
3. Write more and better documentation of the change and how people that do NOT want the index should adjust their code (aka override `Meta.indexes`) to avoid getting them
4. Write a short details explanation of the change for in the changes file with a link to documentation
5. We bumb the major version number of Django-Extensions with the release of this change

(Please note that I will be traveling a lot this month, so appologies in advance for reviews and merging PR's to be a bit slow)

YPCrumble:
@trbs thanks for your thoughtful response! One thing to add.

I think the reason people would assume the indexes are in place is because `TimeStampedModel` [sorts on the `modified` and `created` fields by default]. I believe means that for most users the index will be of benefit any time they query a large number of models unless they've explicitly removed this `Meta` ordering on their `TimeStampedModel`.

Your suggestion for managing the change makes perfect sense. @anudeepsamaiya let me know if you want any help on implementing the changes? I have some time end of next week but not before then.

PR #1434, opened 2019-10-25
[build] Cleanup tox / travis config for py38 and django 3.0

Djailla:
Update the tox and travis configuration to match and to include Python 3.8 and Django 3.0


Djailla:
Hello @trbs you might be interested by this PR following https://github.com/django-extensions/django-extensions/pull/1437

Thanks !

PR #1436, opened 2019-10-25
[compat] Remove py34, deprecated

Djailla:
Python 3.4 is no longer updated a deprecated, should not be supported anymore


trbs:
Please fill in the description !

Python 3.4 is still a supported python version for django until extended support ends at April 2020. 

After that date we can merge this PR.

PR #1441, opened 2019-10-31
Add traceback logging when print_sql is enabled.

Namingwaysway:
This has been an ongoing issue when debugging where slow or redundant database calls were happening in a complex django project.  Having quick access to the traceback here makes identifying and speeding up the removed code a *lot* easier.


trbs:
Thanks !

PR #1442, opened 2019-11-06
Don't pass encoding to json.loads for Python 3.8

novocaine:
This parameter is ignored and deprecated, and passing it triggers this warning:

`DeprecationWarning: 'encoding' is ignored and deprecated. It will be removed in Python 3.9`

### DIFF
--- a/django_extensions/db/fields/json.py
+++ b/django_extensions/db/fields/json.py
@@ -27,8 +27,7 @@ def dumps(value):
 
 def loads(txt):
     value = json.loads(
-        txt,
-        encoding=settings.DEFAULT_CHARSET
+        txt
     )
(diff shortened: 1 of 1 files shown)


novocaine:
sorry, It looks like there are a lot of CI failures unrelated to my change?

trbs:
Since I don't know if people are currently relying on this, it's probably better not to remove this outright but also follow a bit of deprecation schedule ?

So lets add a warning and set a deprecation schedule for removing.

trbs:
Thanks for your help on this !

I've merged the other PR which (temporarily) keeps backwards compatibility support for PY2 until we remove that completely :-)

PR #1446, opened 2019-11-07
added mysql connector django to the mysql engines

### LINKED ISSUE #1445: manage.py command reset_db doesn't work with mysql.connector.django
It says that there is an unknown engine when trying to reset the database. I recommend adding 

```
'mysql.connector.django',
```
 to line 15 of the settings.py of django_extensions

ethanhowell:
Closes #1445


trbs:
Thanks !

Issue #1384, opened 2019-06-13
AutoSlugField: support Django 2.2 UniqueConstraint.condition


thomas-riccardi:
For `AutoSlugField` the code currently seems to only support the old `unique_together`.
Since django 2.2, there is an additional way to specify unique constraints on multiple fields: now with constraint: https://docs.djangoproject.com/en/2.2/ref/models/constraints/#condition

Relevant code:
https://github.com/django-extensions/django-extensions/blob/9bce928111b9bbc82d732a59f7243cdde8d59392/django_extensions/db/fields/__init__.py#L62-L71

yeojin-dev:
May I try to solve this problem? @trbs

trbs:
@yeojin-dev yes please 👍

anudeepsamaiya:
Is a PR created for this? If not can I pick this up?

yeojin-dev:
@anudeepsamaiya Now I'm working on this. :)

PR #1449, opened 2019-11-24
Support Django 2.2 UniqueConstraint.condition 

yeojin-dev:
Issue #1384

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -68,11 +68,13 @@ class UniqueFieldMixin(object):
 
         # for support django 2.2+
         query = Q()
-        for constraint in getattr(model_instance._meta, 'constraints', None):
-            query &= Q(
-                constraint.condition,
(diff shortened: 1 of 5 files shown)


yeojin-dev:
The code I've worked on cannot pass the test in mysql environment only. I don't know why. Do you have any idea why? @trbs

trbs:
> The code I've worked on cannot pass the test in mysql environment only. I don't know why. Do you have any idea why? @trbs

https://docs.djangoproject.com/en/3.0/ref/models/indexes/#django.db.models.Index.condition

Seems MySQL does not support this at all.
Looks like its safe to also skip the tests if MySQL is being used.
(Wondering if we should raise an exception to let the user know it's not going to work as expected, however Django itself also seems to silently ignore the condition.)

yeojin-dev:
Thanks. I fixed it. :) @trbs

trbs:
Thanks !

PR #1450, opened 2019-11-24
Mention required change in settings.py on main page

treszkai:
It took me quite a while to realize that the installation instructions on the main page are only partial, and they continue in another page. Mentioning the `pip` command alone is worse than simply pointing the user to the installation instructions page. This way the instructions are complete, and allows the user to have a quick start.


trbs:
Could you please fix the build error as reported by travis ?

treszkai:
Fixed!

trbs:
Awesome thanks !

PR #1455, opened 2019-12-13
Restore Django 3.0 classifier

adamchainz:
Revert 0f0ca99c8d5f3e34399b2f6a6f920167376d7fb9 because the classifier is now available on PyPI.


trbs:
Thank you for the sharp eye :-)

PR #1458, opened 2020-01-02
Fix AutoSlugField and add test cases

yeojin-dev:
Issue #1384
This is a fix of #1149 PR. (Yes, it's mine.)
I found something new while reading the code. I think a code for `UniqueConstraints.condition` is redundant if `find_unique` will provide a unique slug. It's enough to make `UniqueConstraints.fields` available.


trbs:
Thanks !

Issue #865, opened 2016-04-04
shell_plus notebook settings Import error in subdirectory


crearc:
I have an issue where when I run `python manage.py shell_plus --notebook` I get an error where it can't find my Django settings when I navigate and create a notebook in a subdirectory of the project:

```
WARNING: Attempting to work in a virtualenv. If you encounter problems, please install IPython inside the virtualenv.
[IPKernelApp] WARNING | Error in loading extension: django_extensions.management.notebook_extension
Check your config files in /home/eric/.ipython/profile_default
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/IPython/core/shellapp.py", line 269, in init_extensions
    self.shell.extension_manager.load_extension(ext)
  File "/usr/local/lib/python2.7/dist-packages/IPython/core/extensions.py", line 86, in load_extension
    if self._call_load_ipython_extension(mod):
  File "/usr/local/lib/python2.7/dist-packages/IPython/core/extensions.py", line 133, in _call_load_ipython_extension
    mod.load_ipython_extension(self.shell)
  File "/usr/local/lib/python2.7/dist-packages/django_extensions/management/notebook_extension.py", line 5, in load_ipython_extension
    style=no_style())
  File "/usr/local/lib/python2.7/dist-packages/django_extensions/management/shells.py", line 102, in import_objects
    setup()
  File "/usr/local/lib/python2.7/dist-packages/django/__init__.py", line 17, in setup
    configure_logging(settings.LOGGING_CONFIG, settings.LOGGING)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 48, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 44, in _setup
    self._wrapped = Settings(settings_module)
  File "/usr/local/lib/python2.7/dist-packages/django/conf/__init__.py", line 92, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
ImportError: No module named MY_DJANGO_APP.settings.common
```

It works when I create a python notebook from the root directory of my project, but not from any subdirectory. 

Any suggestions?

chid:
If you haven't resolved the issue you can start the notebook shell from the subdirectory where your notebooks are.

crearc:
[...]

Issue #297, opened 2013-02-01
With ipython 0.13.1, shell_plus --notebook does auto-import models


seanbell:
The django extensions do not load properly with ipython 0.13.1 (current version in `pip`):

    [IPKernelApp] Loading IPython extension: django_extensions.management.notebook_extension
    [IPKernelApp] Error in loading extension: django_extensions.management.notebook_extension

It works under ipython 0.12.

My `settings.py`:

    IPYTHON_ARGUMENTS = [
        '--ext', 'django_extensions.management.notebook_extension',
        '--profile=nbserver',
        '--debug'
    ]

All output:

    $ ./manage.py shell_plus --notebook
    [NotebookApp] Config changed:
    [NotebookApp] {'Application': {'log_level': 10}, 'BaseIPythonApplication': {'profile': u'nbserver'}, 'IPKernelApp': {'extra_extension': u'django_extensions.management.notebook_extension'}}
    [NotebookApp] Using existing profile dir: u'/home/ubuntu/.ipython/profile_nbserver'
    [NotebookApp] Searching path [u'/srv/labelmaterial/server', u'/home/ubuntu/.ipython/profile_nbserver'] for config files
    [NotebookApp] Attempting to load config file: ipython_config.py
    [NotebookApp] Loaded config file: /home/ubuntu/.ipython/profile_nbserver/ipython_config.py
    [NotebookApp] Attempting to load config file: ipython_notebook_config.py
    [NotebookApp] Loaded config file: /home/ubuntu/.ipython/profile_nbserver/ipython_notebook_config.py
    [NotebookApp] Config changed:
    [NotebookApp] {'ZMQInteractiveShell': {'colors': 'Linux'}, 'BaseIPythonApplication': {'profile': u'nbserver'}, 'ProfileDir': {}, 'Application': {'log_level': 10}, 'IPKernelApp': {'pylab': 'inline', 'extra_extension': u'django_extensions.management.notebook_extension'}, 'NotebookApp': {'enable_mathjax': True, 'ipython_dir': u'/home/ubuntu/.ipython', 'open_browser': False, 'ip': '*', 'password': u'removed for security', 'port': 9999}, 'NotebookManager': {'notebook_dir': '/vol/labelmaterial/data/ipython-notebooks'}}
    [NotebookApp] Serving notebooks from /vol/labelmaterial/data/ipython-notebooks
    [NotebookApp] Adding cluster profile 'nbserver'
[...]

PR #1462, opened 2020-01-16
Always add the manage.py directory to PYTHONPATH

terencehonles:
Checking for the manage.py directory being the current directory is not actually needed. If the path ends up being included twice it isn't a very big issue (the first match still wins), but if you want to store notebooks in a directory other than the root of the mange.py directory you currently have to use relative imports to achieve this. This change fixes that so you can always refer to your django package as if it was an installed package.

Fixes: #865


trbs:
FYI: I'm not entirely sure what the reason behind this check was... I could trace it back to https://github.com/django-extensions/django-extensions/issues/297

trbs:
Thanks !

PR #1463, opened 2020-01-17
Add utils create link between M2M fields and ids fields

squallcs12:
```python
Class Author(models.Model):
    name = models.CharField(max_length=80)


class Book(models.Model):
    authors = models.ManyToManyField(Author)
    author_ids = ArrayField(models.IntegerField(), default=list)


make_link_ids_and_many_to_many(Book, 'authors', 'author_ids')
```

### DIFF
--- a/django_extensions/utils/model_cache_ids_fields.py
+++ b/django_extensions/utils/model_cache_ids_fields.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from django.db.models.signals import m2m_changed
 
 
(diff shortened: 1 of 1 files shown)


trbs:
I am not sure if `Django-Extensions` is the right place for this at the moment.

From what I gather this denormalizes the ManyToManyField ?

If so why not instead do something like:

```
from django.contrib.postgres.aggregates.general import ArrayAgg

qs = Book.objects.annotate(author_ids=ArrayAgg('authors__id'))
```

By the time performance becomes an issue with the above I think a generic solution provided by Django-Extensions probably also does not work for the users. (eg: https://github.com/django-extensions/django-extensions/pull/1463/files#diff-1f7ca8594773eb82ab43e10074b170beR11 will not scale as well)

Besides the above questions/feedback; For it to be excepted in Django Extensions it would also need to have documentation, tests, be generic enough for multiple uses and use-cases and be actively maintained.

PR #1464, opened 2020-01-17
Add zh-Hans locale

daimon99:


daimon99:
It is not my code bug. It should be something wrong with the tox testsuit.

trbs:
Thanks !

PR #1476, opened 2020-02-06
I added automatic url tests for all urls of an app

AndreasDickow:
I modified show_urls to test server responses of all urls of a django app

./manage.py test_urls iterates over all urls and returns patterns + url response codes 
at the end the coverage of urls producing no error code is shown, it takes optional arguments --base defaults to 127.0.0.1:8000/ and --mockpk defaults to "1" which is inserted in <placeholder> fields

test_urls requires a running server instance


trbs:
Thanks !

I like the idea, but I'm a bit worried for the code duplication it creates...
Do you think you can split out the common code between show_urls and test_urls ?

Issue #1475, opened 2020-02-05
ImportError when Werkzeug is installed


jameswilliams1:
Running the following code with the latest Wekzeug installed now gives an ImportError (I believe this should now be `from werkzeug.debug import DebuggedApplication`.
https://github.com/django-extensions/django-extensions/blob/0ad61ae76b121344ccd2ad4bbf4af4e4b5b8f70e/django_extensions/management/commands/runserver_plus.py#L222

```
ImportError: cannot import name 'DebuggedApplication' from 'werkzeug' (/usr/local/lib/python3.7/site-packages/werkzeug/__init__.py)
```

Not checked the other imports but a few are at least deprecated and may also be removed soon.

jameswilliams1:
Update: I checked though Werkzeugs code and looks like in the move from 0.16.0->0.16.1 they dropped deprecated top level imports even though it says should have only been done in version 1.0. Raised an issue there as it is technically their problem but I doubt they will patch to re add deprecated code that is due to removed anyway so I guess it needs fixing here.

isolationism:
Yay: Werkzeug 1.0.0 has finally dropped (after only what, a little over a decade?)
Boo: Still broken as per above.

```
$ pip list | grep -i Werkzeug
Werkzeug                  1.0.0     
$ ./manage.py runserver_plus 0.0.0.0:8000
CommandError: Werkzeug is required to use runserver_plus.
Please visit http://werkzeug.pocoo.org/ or install via pip. (pip install Werkzeug)
```

jugmac00:
I created #1477 to fix the import error.

trbs:
Thanks !

mawais78:
> Yay: Werkzeug 1.0.0 has finally dropped (after only what, a little over a decade?)
> Boo: Still broken as per above.
> 
> ```
> $ pip list | grep -i Werkzeug
> Werkzeug                  1.0.0     
> $ ./manage.py runserver_plus 0.0.0.0:8000
> CommandError: Werkzeug is required to use runserver_plus.
> Please visit http://werkzeug.pocoo.org/ or install via pip. (pip install Werkzeug)
> ```

Did anyone find solution of this?

mawais78:
> > Yay: Werkzeug 1.0.0 has finally dropped (after only what, a little over a decade?)
> > Boo: Still broken as per above.
> > ```
> > $ pip list | grep -i Werkzeug
> > Werkzeug                  1.0.0     
[...]

PR #1477, opened 2020-02-06
Fix broken import

jugmac00:
Werkzeug removed a deprecated way to import DebuggedApplication.

The correct import is now:
`from werkzeug.debug import DebuggedApplication`

modified:   django_extensions/management/commands/runserver_plus.py

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -219,7 +219,8 @@ class Command(BaseCommand):
 
     def inner_run(self, options):
         try:
-            from werkzeug import run_simple, DebuggedApplication
+            from werkzeug import run_simple
+            from werkzeug.debug import DebuggedApplication
             from werkzeug.serving import WSGIRequestHandler as _WSGIRequestHandler
 
             # Set colored output


jugmac00:
This fixes #1475

monokrome:
Seems like this bug came in when Werkzeug hit v1.0.0 as a breaking API change =^.^=

trbs:
seems travis is acting up again :-)

trbs:
how backwards compatible is this change ?

or should it be ?
```
try:
    from werkzeug.debug import DebuggedApplication
except ImportError:
    from werkzeug import DebuggedApplication
```

jugmac00:
@trbs  I just had a look at Werkzeug 0.14 (back from 2017), and even there the "new" import works fine.

jameswilliams1:
Just a thought, looks like run_simple has also moved now to werkzeug.serving, possibly a good idea to change the import as well now so there's no need later when they drop that as well?

jugmac00:
@jameswilliams1 Yesterday, I tried to also take care of `run_simple`, but then the test suite just hanged - no error, but it hanged. As I did not want to take any risk, I left that out.

I just tried it again - it is reproducible under Python 3.6 / Ubuntu 18.04, and the current head of django_extensions. The test suite hangs at ` tests/management/commands/test_runserver_plus.py ` only when I change the import from werkzeug to werkzeug.serving - ... for unknown reasons. So at least I won't touch that.

Btw I am neither involved with this project here nor a heavy Flask/Werkzeug user - I just tried to fix some imports in other projects yesterday and @davidsm asked me to to a PR here, too.

So, unfortunately someone else has to look into the `run_simple` import.

Attached is a screen where my test suite hangs with `run_simple` imported from werkzeug.serving

Screenshot from 2020-02-07 13-03-58

jameswilliams1:
Strange, lucky it still works for now then. Thanks very much for the fix.

trbs:
Thanks guys, will merge and do a new release to get this out of the way.

I'll try to find some time and look at the `run_simple` issue.

PR #1492, opened 2020-02-24
Replace deprecated `force_text()` with `force_str()`

Tenzer:
`force_text()` has been deprecated since Django 3.0 and was an alias for
`force_str()` since Django 2.0:
https://docs.djangoproject.com/en/dev/releases/3.0/#deprecated-features-3-0

It's set to be removed in Django 4.0.

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -27,7 +27,7 @@ from django.db.models import DateTimeField, CharField, SlugField, Q
 from django.db.models.constants import LOOKUP_SEP
 from django.template.defaultfilters import slugify
 from django.utils.crypto import get_random_string
-from django.utils.encoding import force_text
+from django.utils.encoding import force_str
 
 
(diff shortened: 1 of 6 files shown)


Tenzer:
I guess somewhat expectedly this will change the behaviour of when run with Django 1.11. I can see it's scheduled to be supported by the Django team until April 2020: https://www.djangoproject.com/download/#supported-versions.

I don't know if it would make sense to make a special case for Django 1.11 and add extra complexity to the code in order to use `force_text()` in Django 1.11, while using `force_str()` for the rest?

trbs:
Thinking we could wait a little bit with merging this PR until `Django 1.11`.
After it is fully unsupported we don't have to worry about special casing it anymore :-)

trbs:
(Ofcourse if you want to to do work for the update & adding support for Django 1.11 I'm more then happy to merge it before then. We can also remove the special case together with the other cleanup changes after Django 1.11 reaches EOF.)

clarkmoody:
It looks like extended support for Django 1.11 has ended.

Tenzer:
I've rebased the changes so the PR is mergeable again.

trbs:
Awesome thank you !

PR #1504, opened 2020-03-25
docs: Fix simple typo, writeable -> writable

timgates42:
There is a small typo in django_extensions/management/utils.py, tests/management/commands/test_graph_models.py.

Should read `writable` rather than `writeable`.


trbs:
Thanks !

PR #1508, opened 2020-04-02
Support multiple DBs in the debug_cursor code

JBKahn:
I work on a django app with multilple DBs and I've been using this hack for a while in my site-packages folder and thought I'd push it upstream.

If you'd like me to add a test, I'll need some help with that. The basic case of 1 DB is tested but not explicitly, just that `--print-sql` prints the SQL rather than which DB it is from.

```
In [2]: User.objects.exists()
SELECT (1) AS "a"
  FROM "wave_auth_user"
 LIMIT 1

Execution time: 0.026820s [Database: default]
Out[2]: False

In [4]: from businesses.models import Business

In [5]: Business.objects.using('shard_0001').exists()
SELECT (1) AS "a"
  FROM "businesses_business"
 LIMIT 1

Execution time: 0.008085s [Database: shard_0001]
Out[5]: False
```

### DIFF
--- a/django_extensions/management/debug_cursor.py
+++ b/django_extensions/management/debug_cursor.py
@@ -109,6 +109,6 @@ def monkey_patch_cursordebugwrapper(print_sql=None, print_sql_location=False, tr
         if postgresql_base:
             postgresql_base.CursorDebugWrapper = _PostgreSQLCursorDebugWrapper
 
-        if connection:
+        if connections:
             for connection_name in connections:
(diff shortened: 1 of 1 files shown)


trbs:
Thanks !

Tests would be preferred. 

@tjaartvanderWalt @mwarkentin would you be able to help with this ?

trbs:
If somebody could still help with tests that would be much appreciated !
For now I will merge it anyways.

Thanks @JBKahn !

PR #1509, opened 2020-04-08
update shell_plus notebook to allow overriding kernelspec handling

### LINKED ISSUE #1378: Jupyter Kernel installed under Library/Jupyter  rather than virtual environment
I have a Django project that uses pipenv to manage its virtual environment, in which I have installed Jupyter and "django-extensions" and run `/manage.py shell_plus --notebook`.

It seems the "django_extensions" jupyter kernel gets installed in `Library/Jupyter/kernerls`, making is accessible to other jupyter projects. By contrast, it seems like the python3 kernel is installed within the virtual environment.

Here is what I see when running `jupyter kernelspec list`, inside my project `cc-dashboard` with its respective pipenv activated.

```bash
jupyter kernelspec list

Available kernels:
  django_extensions    /Users/dustinmichels/Library/Jupyter/kernels/django_extensions
  python3              /Users/dustinmichels/.local/share/virtualenvs/cc-dashboard-15QDmC0q/bin/../share/jupyter/kernels/python3
```

Is there a way to get the kernel to install itself under the virtual environment's `share/jupyter` folder instead? Thanks.

## Specs

- **django** -- version: 2.2.1
- **django-extensions** -- version: 2.1.7
- **jupyter** -- version: 4.4.0
- **python** -- version: 3.7.3
- **pipenv** -- version 2018.11.26
- **OS** -- MacOS Mojave (10.14.5)

### LINKED ISSUE #1422: Multiple Settings files
I'm wondering, if my app has more than one settings file, how can I add multiple Django-Shell kernels?

I didn't find this is docs and at the first glance in source code it is  impossible to do as well.

Am I miss something?

Thanks

terencehonles:
hoists/refactors `install_kernel_spec` from being a closure to a method `generate_kernel_specs`, and splits out finding the base kernel spec into method `load_base_kernel_spec` so it is easier to override `generate_kernel_specs` with a new command and install one or more kernel specs.

fixes: #1026, fixes #1378, fixes #1422

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -171,6 +171,46 @@ class Command(BaseCommand):
             start_kernel(**kwargs)
         return run_kernel
 
+    def load_base_kernel_spec(self, app):
+        """Finds and returns the base Python kernelspec to extend from."""
+        ksm = app.kernel_spec_manager
+        try_spec_names = getattr(settings, 'NOTEBOOK_KERNEL_SPEC_NAMES', [
+            'python3' if PY3 else 'python2',
+            'python',
+        ])
+
+        if isinstance(try_spec_names, six.string_types):
+            try_spec_names = [try_spec_names]
+
+        ks = None
+        for spec_name in try_spec_names:
+            try:
+                ks = ksm.get_kernel_spec(spec_name)
+                break
+            except Exception:
+                continue
+        if not ks:
+            raise CommandError("No notebook (Python) kernel specs found. Tried %r" % try_spec_names)
+
+        return ks
+
+    def generate_kernel_specs(self, app, ipython_arguments):
+        """Generate an IPython >= 3.0 kernelspec that loads django extensions"""
+        ks = self.load_base_kernel_spec(app)
+        ks.argv.extend(ipython_arguments)
+        ks.display_name = getattr(settings, 'IPYTHON_KERNEL_DISPLAY_NAME', "Django Shell-Plus")
+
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/shell_plus.py
```diff
@@ -171,6 +171,52 @@ def run_kernel():
...
+
+        manage_py_dir, manage_py = os.path.split(os.path.realpath(sys.argv[0]))
+
+        if manage_py == 'manage.py' and os.path.isdir(manage_py_dir):
+            pythonpath = ks.env.get('PYTHONPATH', os.environ.get('PYTHONPATH', ''))
+            pythonpath = pythonpath.split(os.pathsep)
+            if manage_py_dir not in pythonpath:
+                pythonpath.append(manage_py_dir)
+
+            ks.env['PYTHONPATH'] = os.pathsep.join(filter(None, pythonpath))
```
terencehonles, line 212: This functionality could be moved to `load_base_kernel_spec` but I avoided mutating the kernel spec and left everything in this method. This meant I had to duplicate this logic in my `install_kernel_spec` override, but there may(?) be a case where you don't want `PYTHONPATH` touched?

trbs: Maybe split this out in yet-another-method like ? 

At least that makes it easier to include it in a custom `install_kernel_spec` or override this particular behavior if needed...

terencehonles: I refactored my refactoring and now I can just call `super().generate_kernel_specs(...)` and modify the result to create the sibling kernel specs I was trying to create.

trbs:
Thanks !

PR #1510, opened 2020-04-09
shell_plus: fix handling of extra_args in --notebook

terencehonles:
Updated to be based off of #1511 which is moving around the `run_notebook` code.

In `run_notebook` after `get_ipython_arguments` is called and the result is modified, but since both `get_ipython_arguments` and `get_notebook_arguments` refer to the same list this produces an invalid command line.

Simply returning new lists does not fix the problem because both the notebook and the kernel fail on invalid flags. This fix handles extra_args explicitly to ensure the command line is built correctly.

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -171,6 +171,46 @@ class Command(BaseCommand):
             start_kernel(**kwargs)
         return run_kernel
 
+    def load_base_kernel_spec(self, app):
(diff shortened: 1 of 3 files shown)


trbs:
Think this is duplicate from the other PR's ?

terencehonles:
@trbs I can merge it into the other PRs, but it is slightly different and is only commit 1aebcfdbdc9d5f421ee3ca8254760938b17980ee

PR #1515, opened 2020-04-13
update shell_plus to remove boilerplate code to add new shells

terencehonles:
- Change the usage of "shell(s)" in the command handler to use "runner" instead. This change more accurately reflects that some of the runners are not your typical shells (``--notebook`` and ``--lab`` start notebook servers)
- The command handler now started one of the decorated "runners" which allow the runner to define the flag(s) it responds to and its name. This will make it simpler to not have to add the runner to the argument parser *and* also the previous "shells" list (while ensuring someone doesn't place the new shell at the top of the list which alters the "auto shell" behavior)
- Runners are all registered in a mutually exclusive argument parser group so you cannot specify ``--plain`` and ``--notebook`` at the same time and have the priority rely on the ``if/else`` chain which switches between the shells.
- Change the exit code to an error code if a shell could not be started

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
+import inspect
 import os
-import six
 import sys
 import traceback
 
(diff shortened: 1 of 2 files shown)


trbs:
Sorry for the late response, I would love to merge the PR.
However I also committed some work to start removing Python 2 and Django 1.11 support.

Could you please update the PR so we can merge it into mainline ?

terencehonles:
@trbs not sure if you noticed, but I did update the PR to not have any conflicts.

trbs:
@terencehonles thanks !

Regretfully somehow Github and Travis broke and the results does not show up anymore here...

trbs:
Travis build reporting issues got fixed :-)

Thanks for your work !

PR #1518, opened 2020-04-19
Migrate from keyczar

### LINKED ISSUE #1497: Migrating from keyczar
we are updating an older application which is using this encrypted field with keyczar. Reading #1359 it's obvious this was all deprecated a long time ago. 
I am wondering if anyone has worked on some upgrade path to have encrypted fields using another library like
https://github.com/luojilab/django-mirage-field
or
https://github.com/georgemarshall/django-cryptography
or something else. We are looking at how keyczar was encrypting the data to see if we could do it the same with another package. But were thinking it may be easier to run a batch job to decrypt the fields using the keyczar method and re-encrypt them using a new method. But the old fields probably cant be decrypted.
So i was hoping to see if anyone else had any input on this.

yeojin-dev:
Resolves #1497 

I have just turned keyczar into cryptography by using symmetric encryption algorithms.


django_extensions/db/fields/encrypted.py
```diff
@@ -24,30 +28,19 @@ class BaseEncryptedField(models.Field):
...
-            "all code we highly recommend to stop using potentially unsafe abandonware directly "
-            "these versions of encrypted fields will be removed soon but it's highly likely their "
-            "names will be reused. Please see github issues/1359 for discussion on possible "
-            "replacement fields",
-            RemovedInNextVersionWarning,
-            stacklevel=2,
-        )
+        if not getattr(settings, 'CRYPTOGRAPHY_ENCRYPT_ALGORITHM', None):
+            raise ImproperlyConfigured('You must set the settings.CRYPTOGRAPHY_ENCRYPT_ALGORITHM '
+                                       'setting to your cryptography keys directory.')
```
terencehonles, line 33: This message doesn't seem to match the setting

django_extensions/db/fields/encrypted.py
```diff
@@ -24,30 +28,19 @@ class BaseEncryptedField(models.Field):
...
-            raise ImproperlyConfigured(
-                "Please do not use these fields. Since Keyczar is deprecated and authors deleted "
-                "all code we highly recommend to stop using potentially unsafe abandonware directly "
-                "these versions of encrypted fields will be removed soon but it's highly likely their "
-                "names will be reused. Please see github issues/1359 for discussion on possible "
-                "replacement fields",
-            )
+        if not getattr(settings, 'CRYPTOGRAPHY_ENCRYPT_KEY', None):
+            raise ImproperlyConfigured('You must set the settings.CRYPTOGRAPHY_ENCRYPT_KEY '
+                                       'setting to your cryptography keys directory.')
```
terencehonles, line 37: This message doesn't seem to match the setting

django_extensions/db/fields/encrypted.py
trbs: We should also name this a bit different / less generic.

I would also don't mind using a dict in settings.py for configuration, specially if there is more things to parameterize.

For example (but by no means an official recommendation :-) : 
```
DJANGO_EXTENSIONS_ENCRYPTION_FIELDS = {
  "CRYPTOGRAPHY_ENCRYPT_ALGORITHM": "AES",
  "CRYPTOGRAPHY_ENCRYPT_KEY": os.urandom(16),
  "PBKDF2HMAC_ALGORITHM": "SHA256",
  "PBKDF2HMAC_LENGTH": 32,
  "SALT": os.urandom(16),
}

# or 

DJANGO_EXTENSIONS_ENCRYPTION_FIELDS_ALGORITHM = "AES"
```

trbs:
Think @terencehonles makes some good points.

As this PR replaces the functionality of the fields (vs adding new encryption fields) I think we do need to support some way of migration for our users. In the form of some documentation and some commands (or scripts we can link to if we don't want them to pollute the commands list) to help people migrate to the new fields.

Sorry for some of the conflicts, as we start removing support for unsupported Django and Python versions this might break a couple of PR's, my apologies for the trouble but please rebase.

yeojin-dev:
@terencehonles @trbs Thank you. :)

django_extensions/db/fields/encrypted.py
```diff
@@ -23,30 +23,11 @@ class BaseEncryptedField(models.Field):
     prefix = 'enc_str:::'
```
trbs, line 23: Should we also use a different prefix ? such that we can detect old keyczar values ?

trbs:
Sorry for the slow progress on this. It's a significant change and while we are due for a major release of Django Extensions (thus making it worthwhile to include this) I also want to get it right :-)

Alternative option would be that we first completely drop `keyzcar` and the fields from the next release. This would also force people to understand that it really is not supported anymore. But I guess the drawback would be that any migration plan and/or documentation (something I would love to include) would be much harder...

yeojin-dev:
My suggestion is to use a completely new class name. What about names like `CryptoField`? While the existing `EncryptedField` class will remain for a certain period of time, providing a guide and a django command to transfer data from the `EncryptedField` to the `CryptoField` will help users migrate. I will also write the code related to it. @trbs

yeojin-dev:
What is your current opinion? If you want to drop `EncryptedField` first, I would like to make a PR related to it first. :) @trbs

trbs:
Let's do that... keyczar was deprecated a long long time ago.... and since the next version is going to be a major version bump I would love to get as much breaking things in there as possible :-)

Then we can start adding things later... would you mind creating the PR to completely drop `EncryptedField` first ?

django_extensions/db/fields/encrypted.py
```diff
@@ -0,0 +1,101 @@
...
+        super().__init__(*args, **kwargs)
+
+    def calculate_crypt_max_length(self, unencrypted_length):
+        # TODO: Re-examine if this logic will actually make a large-enough
+        # max-length for unicode strings that have non-ascii characters in them.
+        # For PostGreSQL we might as well always use textfield since there is little
+        # difference (except for length checking) between varchar and text in PG.
+        test_bytes = ('x' * unencrypted_length).encode()
+        encrypted_data = base64.urlsafe_b64encode(self.fernet.encrypt(test_bytes)).decode()
+        return len(encrypted_data)
```
terencehonles, line 45: I think allowing this seems a little odd, you may want to make the field validate the decrypted data is a specific length **before** encrypting rather than encrypting a test string and testing the encrypted length. If you still feel that it would be better to check the max length of the encrypted string, I would probably cache this value rather than run an extra encryption per `__init__`. Since it's on the model maybe I'm prematurely optimizing, but I think this would be pretty easy to do with and since it's almost what you have hear it's worth just caching the values:

```python
encrypted_lengths = {}
def calculate_crypt_max_length(unencrypted_length):
    if unencrypted_length not in encrypted_lengths:
        test_bytes = ('x' * unencrypted_length).encode()
        encrypted_data = base64.urlsafe_b64encode(self.fernet.encrypt(test_bytes)).decode()
        encrypted_lengths[unencrypted_length] = len(encrypted_data)

    return encrypted_lengths[unencrypted_length]
```

django_extensions/db/fields/encrypted.py
```diff
@@ -0,0 +1,101 @@
...
+        # Truncated encrypted content is unreadable,
+        # so truncate before encryption
+        max_length = self.unencrypted_length
+        if max_length and len(value) > max_length:
+            warnings.warn("Truncating field %s from %d to %d bytes" % (
+                self.name, len(value), max_length), EncryptionWarning
+            )
+            value = value[:max_length]
+
+        encrypted_data = self.fernet.encrypt(value.encode())
```
terencehonles, line 70: Should this be relying on UTF-8 encoding? and also storing the data as base64 encoded binary? I'm not familiar with keyczar, but I _think_ this should probably be using a `BinaryField` and not base64 encoding, for the input if this is a `EncryptedCharField` you may want to have the user be allowed to pass the encoding (so this base class would have a `self.encoding` which could default to UTF-8), but I would think you might want to also provide a `EncryptedBinaryField` which has `self.encoding` of `None` which would have this base class not try to decode the data and just return the bytes directly.

trbs: Good questions... my first though is to have multiple variants for this...

I would indeed want the data stored in `BinaryField` most of the time.... (since that makes a lot of sense to me and cuts down on overhead)...

When I do want so much portability that I see the need for using some kind of character field I would actually think 99/100 for me the size (max_length) does not matter.. Since I almost exclusively use PostgreSQL there is little difference between `varchar` and `text` for me.

However I can understand that there are also use-cases for people out there wanting  a `CharField(max_length=123)`...

On another note... would it also not make sense to start storing these things like we do with hashed passwords ?

Much like `User.password` uses `pbkdf2_sha256$180000$fgB0GTcbB2Rg$rpkfa5oqz4RHqZYS+Bsm3ZPNSoVUWU3mWitXI1fxuto=`

Might we not want to store something like:

`fernet$1$gAAAAABfIdWwisKpNayqvyyrK0TmsUjf2vZWTgpEupiGipi2t2V-KZFLXsekO5Bfh2TaNCeyFtKW6_NwP1u0dEH3XhlpQ9rAYA==`

Where first value is encryption type `fernet`, the second is flag / version field for forward compatibility and the third value is the encrypted string itself (here base64 encoded).

django_extensions/db/fields/encrypted.py
trbs: As said in the other comment... I can well imaging that for most use-cases having a limit on the database is not a big deal... then it would be enough to check the length for validation purposes of the unencrypted value.

Also if the documentation clearly states that encrypted values are longer (prefix + whatever the encryption length expansion is) then I think that would be good enough as well... Then people are responsible themselves for dealing with this... assuming most people can live with a TextField implementation or setting something like `max_length=128, db_max_length=256` and the onces that need more should be able to easily subclass the `Field` and implement there own behavior... 

(unless somebody has a easy to implement, easy to understand, side effect free solution for this :-)

yeojin-dev:
Close because it is a stale PR that has not been completed.

PR #1523, opened 2020-05-06
Use logging to record runjobs errors

pdewacht:
This PR changes the runjobs command to use the logging framework to report errors. This way, cron errors can be handled the same way as other errors. In our case, that means the errors will be recorded in sentry.io.

The default behavior remains to just write a traceback to stdout.


trbs:
+1 lgtm

trbs:
Thanks !

PR #1525, opened 2020-05-09
Patches to make MySQL sqldiff work (at least on our projects)

xurble:
This Pull Request makes sqldiff (appear to) work for MySQL in our projects.

I am not sure if this is the best way to implement the fixes as I basically copy 2 whole functions from the superclass into the MySQL class and then tweak them.  That leaves them a bit vulnerable to not picking up changes in the superclass but I wasn't sure what else to do.

There may still be issues with MySQL but this eliminates a lot of false positives.


trbs:
Awesome, thanks !

PR #1527, opened 2020-05-13
Fix silent errors when using syncdata

cuchi:
There is an issue while importing broken fixtures (i.e. when there is an invalid UUID).
The fixture fails silently and interrupts the whole process, resulting in a success message `Installed x objects from y fixtures`.

This happens because another exception is thrown when attempting to rollback a failed transaction, which was the case.

### DIFF
--- a/django_extensions/management/commands/syncdata.py
+++ b/django_extensions/management/commands/syncdata.py
@@ -179,7 +179,8 @@ class Command(BaseCommand):
                             except Exception:
                                 import traceback
                                 fixture.close()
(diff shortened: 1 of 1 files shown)


trbs:
The build error should fix itself after rebasing to master.

Would it be possible to have a test case for this ?
And/or maybe have a bit more detailed explanation why using `get_rollback` is the correct fix.

cuchi:
Closed in favor of #1532

trbs:
Thanks for your help and work on this @cuchi !

PR #1529, opened 2020-05-20
Add flag on syncdata to allow removing objects before

cuchi:
This covers the case when the object being upserted breaks a constraint with another existing object. This kind of fixture will only work properly if the extraneous objects get removed first.

### DIFF
--- a/django_extensions/management/commands/syncdata.py
+++ b/django_extensions/management/commands/syncdata.py
@@ -40,6 +40,10 @@ class Command(BaseCommand):
             help='Avoid remove any object from db',
         )
         parser.add_argument(
+            '--remove-before', action='store_true', dest='remove_before', default=False,
+            help='Remove existing objects before inserting and updating new ones',
+        )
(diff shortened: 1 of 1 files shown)


django_extensions/management/commands/syncdata.py
```diff
@@ -157,20 +161,23 @@ def syncdata(self, fixture_labels, options):
                                 print("Installing %s fixture '%s' from %s." % (format, fixture_name, humanize(fixture_dir)))
                             try:
                                 objects_to_keep = {}
-                                objects = serializers.deserialize(format, fixture)
+                                objects = list(serializers.deserialize(format, fixture))
```
trbs, line 164: Won't it be slightly more efficient to use the `objects_to_keep` in the subsequent runs instead of forcing the materialization of `objects` here ?

trbs:
For now I will merge the PR, please still look at the comment and do a second merge request if you can.

PR #1531, opened 2020-05-24
Fixes deprecation on Django 3.1

jannh:
URL routing imports now raise a warning, change needs Django 2.0+

Update documentation that only Django 2.2+ is supported now (According to commit on April 21st)


trbs:
Thanks !

Issue #1215, opened 2018-06-26
TransactionManagementError with syncdata management command


Apreche:
I'm running django-extensions 2.0.7, Django 2.0.5 and Python 3.5.2. 

When I run the command `./manage.py syncdata` I see this error:

```
Traceback (most recent call last):
  File "./manage.py", line 15, in <module>
    execute_from_command_line(sys.argv)
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/core/management/__init__.py", line 371, in execute_from_command_line
    utility.execute()
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/core/management/__init__.py", line 365, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/core/management/base.py", line 335, in execute
    output = self.handle(*args, **options)
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django_extensions/management/utils.py", line 59, in inner
    ret = func(self, *args, **kwargs)
  File "/usr/lib/python3.5/contextlib.py", line 30, in inner
    return func(*args, **kwds)
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django_extensions/management/commands/syncdata.py", line 190, in handle
    transaction.commit()
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/db/transaction.py", line 35, in commit
    get_connection(using).commit()
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/db/backends/base/base.py", line 260, in commit
    self.validate_no_atomic_block()
  File "/home/myuser/.virtualenvs/myproject/lib/python3.5/site-packages/django/db/backends/base/base.py", line 432, in validate_no_atomic_block
    "This is forbidden when an 'atomic' block is active.")
django.db.transaction.TransactionManagementError: This is forbidden when an 'atomic' block is active.
```

trbs:
Please test with ffbda59286fb40c91dd3a7cc1c5b60fae0ac3a8d

trbs:
please reopen if it does not fix the issue.

[...]

PR #1532, opened 2020-05-25
Syncdata silent failure on DeserializationError

MattFisher:
Addresses the extra rollback issue in #1215
https://github.com/django-extensions/django-extensions/issues/1215#issuecomment-593774907

When `syncdata` calls `serializers.deserialize(format, fixture)`, it's possible to get a `DeserializationError`, eg from a `django.core.exceptions.FieldDoesNotExist` if the fixture contains a field that as since been deleted.

This exception gets caught by the `except Exception` block, but calling `transaction.rollback()` causes a `TransactionManagementError: This is forbidden when an 'atomic' block is active.` before the traceback gets printed or the appropriate `SyncDataError` gets raised, and then this gets swallowed by the next `except Exception` block.

Since the whole call to `syncdata` is in an atomic transaction context manager, it's safe to just raise the appropriate exception and let the context manager take care of the rollback.

This PR includes a test to expose the error and removes the call to `transaction.rollback()`.

As an aside, it seems that the exit code is always 0, even if there are errors in the command. I haven't looked at other management commands to see if this is consistent across the project, but is there a reason the exit code isn't non-zero if an error occurs? Could make the commands more useful in CI/automation scenarios.

### DIFF
--- a/django_extensions/management/commands/syncdata.py
+++ b/django_extensions/management/commands/syncdata.py
@@ -186,7 +186,6 @@ class Command(BaseCommand):
                             except Exception:
                                 import traceback
                                 fixture.close()
-                                transaction.rollback()
                                 if show_traceback:
                                     traceback.print_exc()
                                 raise SyncDataError("Problem installing fixture '%s': %s\n" % (full_path, traceback.format_exc()))
--- a/tests/management/commands/fixtures/fixture_with_nonexistent_field.json
+++ b/tests/management/commands/fixtures/fixture_with_nonexistent_field.json
@@ -1,0 +1,21 @@
+[
+  {
+    "pk" : 3,
+    "fields" : {
+      "is_superuser" : false,
+      "first_name" : "John",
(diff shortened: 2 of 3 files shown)


MattFisher:
Hah guess I should have checked current PRs before doing this - didn't see #1527 😆

cuchi:
> Hah guess I should have checked current PRs before doing this - didn't see #1527 :laughing:

After further investigation, I found out my change doesn't work in every case, and since yours has tests, I'm closing #1527 :wink:

trbs:
@MattFisher I would classify returning exit code 0 if a failure happens a bug ! :-)

trbs:
Thanks @MattFisher and @cuchi for your work on this !

trbs:
@MattFisher please check https://github.com/django-extensions/django-extensions/commit/d9777aa5d6682f3f143ec189f4063d7df5aa63b7

Using `CommandError` should now also provide a non-zero exit code on failures.

PR #1535, opened 2020-06-05
Correct grammar for django-extensions

echirchir:
There were some grammar issues with the documentation:
-----BEFORE-----

Django 2.0 now contains similar functionality as ForeignKeyAutocompleteAdmin therefor we are deprecated this extension and high encourage everybody to update to it.
This code will be removed in the near feature when support for Django older then 2.0 is dropped.
-----AFTER---------
Django 2.0 now contains similar functionality as *ForeignKeyAutocompleteAdmin* therefore we are deprecating this extension and highly encouraging everyone to update to it.

----

### DIFF
--- a/docs/admin_extensions.rst
+++ b/docs/admin_extensions.rst
@@ -16,9 +16,9 @@ Current Admin Extensions
 Depreciation
 ------------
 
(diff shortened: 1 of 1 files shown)


trbs:
Thank you !

Issue #1537, opened 2020-06-09
Is there any reason why the fields are not fully included in graph_models attributes?


Gagaro:
I'm trying to make a custom theme for the graph_models command, and the data available is very limited.

Is there a reason why this is so limited (https://github.com/django-extensions/django-extensions/blob/master/django_extensions/management/modelviz.py#L133-L141)?

Why would you think about adding the field itself (something like that https://github.com/Gagaro/django-extensions/commit/cb62f43fe81bcc377d5c3e25af437d00fe84533f)?

Thanks

trbs:
I don't think there is a particular reason :-)

It's a bit messy (in the code) but I don't seen any reason from the top of my head not to include it.

Basically this gives you more information to work with during templating the dot file right ?
I would say make a PR for it :-)

Gagaro:
> Basically this gives you more information to work with during templating the dot file right ?
Yes that's exactly that.

> I would say make a PR for it :-)
Will do, thanks for the fast reply.

PR #1538, opened 2020-06-09
Add field and model in modelviz

Gagaro:
Related to #1537.

I also added `model` as it is the same deal than the fields.


Gagaro:
Because the generation of the data and the retrieval of the data are two distinct things, the best way I found is to drop the field and model values from the data when we get it if is for a JSON export.

trbs:
Thanks !

Issue #1549, opened 2020-07-03
syncdata unable to use natural keys


MattFisher:
Version 3 seems unable to use natural keys in `syncdata` fixtures.

The traceback shows a long string of "During handling of the above exception, another exception occurred", but the important bits are:
```
django_extensions.management.commands.syncdata.SyncDataError: Problem installing fixture '.../myproject/fixtures/base.json': Traceback (most recent call last):
  File ".../python3.6/site-packages/django/core/serializers/base.py", line 288, in deserialize_m2m_values
    values.append(m2m_convert(pk))
  File ".../python3.6/site-packages/django/core/serializers/base.py", line 278, in m2m_convert
    return model._default_manager.db_manager(using).get_by_natural_key(*value).pk
  File ".../python3.6/site-packages/django/contrib/auth/models.py", line 88, in get_by_natural_key
    return self.get(name=name)
  File ".../python3.6/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File ".../python3.6/site-packages/django/db/models/query.py", line 417, in get
    self.model._meta.object_name
django.contrib.auth.models.Group.DoesNotExist: Group matching query does not exist.

...

Traceback (most recent call last):
  File ".../python3.6/site-packages/django_extensions/management/commands/syncdata.py", line 166, in syncdata
    objects = list(serializers.deserialize(format_, fixture))
  File ".../python3.6/site-packages/django/core/serializers/json.py", line 69, in Deserializer
    yield from PythonDeserializer(objects, **options)
  File ".../python3.6/site-packages/django/core/serializers/python.py", line 124, in Deserializer
    raise base.DeserializationError.WithData(e.original_exc, d['model'], d.get('pk'), e.pk)
django.core.serializers.base.DeserializationError: Group matching query does not exist.: (users.user:pk=2) field_value was '['teachers']'
```

I think this is because of the changes made here https://github.com/django-extensions/django-extensions/commit/0edbb6cf31d8851c39a4ea0ed52092b8cc2b60d2#diff-ed4bf9a1cad19a0969a26b755518c4a9R164,  in adding the `remove_before` flag.

In django's `loaddata`, it uses the construct:
```
objects = serializers.deserialize(
    ser_fmt, fixture, using=self.using, ignorenonexistent=self.ignore,
    handle_forward_references=True,
)

for obj in objects:
    # snip
    obj.save(using=self.using)
```
[...]

PR #1551, opened 2020-07-05
Failing test and fixture for syncdata with natural keys

MattFisher:
Failing test for Issue #1549


trbs:
I would love to merge but i am busy at work and likely won't have the time to work on a fix :-(

What we can do if we cannot make a PR for a permanent solution soon is to put one of these tests on expected to fail ( @xfail  ).

And maybe revert the previous change depending on what's the worst problem to keep :-)

camilonova:
@MattFisher Looks like the fixtures are not loading properly. Would you take a look and open a new PR so we can start fresh?

Issue #1524, opened 2020-05-07
[print_settings] Hide secret field values by default


david-libremone:
**User story:** I am a well-intentioned sysadmin. I want to print and share Django settings with my colleague, so that I can investigate a problem without revealing secret values.

**Proposed solution:**

1. Define the following "sensitive" patterns: `SECRET`, `PASSWORD`

2. All settings matching these patterns show a value of `**********`

3. If the user passes the parameter `--show-secrets`, then show the actual values.

trbs:
Very good idea !

We should also allow users to override the sensitive patterns in `settings.py` so people can add their own as needed.

Could you make a PR for this ?
Happy to merge it !

pdewacht:
FYI, Django already has a `get_safe_settings` function which applies the filtering.

luchiago:
Hi, @david-libremone and @trbs. I'm looking for the first time this project and I would like to contribute. Is this still an issue?

trbs:
@luchiago ik believe so.  (I don’t remember merging a PR for this)

You could start by confirming that secrets are still present in the default output. Eg: the value of ‘SECRET_KEY’ or ‘DATABASES[‘default’][‘PASSWORD’]’ settings. Or you can dive straight into the command to see if safe settings (or similar) is being used in the code :-)

If the raw values still exists changing the command to utilize ‘get_safe_settings’ or similar would be much appreciated !

(Would make for an awesome first contribution :-) )

luchiago:
@trbs Thanks. I didn't have time to investigate this issue, but sure I will send a contribution for it.

Akshat1903:
Hi, @david-libremone and @trbs. I'm looking to contribute on this project and as a beginner chose an easy picking. Just wanted to confirm is this still an open issue?

trbs:
Hi @Akshat1903, Yeah this is still an open issue and I think a good one to start with.

There is already an open PR for it which you might also want to take a look at.

(Please understand that it might take awhile for a PR to be reviewed as I do not have a lot of spare time to work on Django Extensions these days. But don't be discouraged ;-) any help is much appreciated !!!)

Akshat1903:
Thanks for informing, i'll start by studying the codebase and then look into the issue.

PR #1553, opened 2020-07-12
Hide secrets

luchiago:
This PR adds a new option during `print settings` call, to hide/show the secrets, replacing them with '*****'. #1524


luchiago:
> please do not run `black` and `isort` in this PR, even though I love both tools it makes reviewing a PR much harder due to unrelated changes.

I can remove the commit where I did, rebase, and send again. Is there another change to make in this PR?

trbs:
> > please do not run `black` and `isort` in this PR, even though I love both tools it makes reviewing a PR much harder due to unrelated changes.
> 
> @trbs I can remove the commit where I did, rebase, and send again. Is there another change to make in this PR?

You can make it into a separate PR for this file and I will merge it.

django_extensions/management/commands/print_settings.py
```diff
@@ -26,34 +31,47 @@ def add_arguments(self, parser):
...
-            default=4,
+            default=DEFAULT_INDENT,
             dest='indent',
             type=int,
             help='Specifies indent level for JSON and YAML'
         )
+        parser.add_argument(
+            '--show-secrets',
+            default=DEFAULT_SECRETS,
+            dest='show secrets',
```
trbs, line 48: please use underscores not spaces

django_extensions/management/commands/print_settings.py
```diff
@@ -84,3 +111,13 @@ def include_attr(attr, settings):
     def print_simple(a_dict):
         for key, value in a_dict.items():
             print('%-40s = %r' % (key, value))
+
+    @staticmethod
+    def safe_settings():
+        try:
+            return SafeExceptionReporterFilter().get_safe_settings()
```
trbs, line 118: same comment as on the other PR :-)

Either the comment below is wrong or the code is... if django < 3.0 does not have the class then the imports at the top will already have raised an ImportError.

Also if this method does not serve any other purpose just deal with this in its entirely at the top of the file and at startup time.

luchiago:
> > > please do not run `black` and `isort` in this PR, even though I love both tools it makes reviewing a PR much harder due to unrelated changes.
> > 
> > 
> > @trbs I can remove the commit where I did, rebase, and send again. Is there another change to make in this PR?
> 
> You can make it into a separate PR for this file and I will merge it.

Sure. I will finish the adjustments here and send this PR

django_extensions/management/commands/print_settings.py
luchiago: @trbs do you mean, have a try-catch block at the top of the file and make an alias for the correct import? Like:
```python
try:
    from django.views.debug import SafeExceptionReporterFilter as settings
except ImportError:
    from django.views.debug import get_safe_settings
```

luchiago: I made this function because Django > 3.0 does exist in this class and inside of it, there's the method. For Django < 3.0 does exist only the method

trbs: Yes, if the problem is an `ImportError` then the above should do otherwise you can so something similar with an `AttributeError`.

But in the latter case at the very least the comment needs to be updated to reflect this.

luchiago: I think we should keep in this way:
```python
try:
    from django.views.debug import SafeExceptionReporterFilter
except ImportError:
    from django.views.debug import get_safe_settings
...
    @staticmethod
    def safe_settings():
        try:
            return SafeExceptionReporterFilter().get_safe_settings()
        except AttributeError:
            # In django < 3.0 does not have this class
            # We have to make a different import
            return get_safe_settings()
```
Is safer, we have only one function to return the safe settings and we can update the comment. How'd you think should be the new comment?

trbs: This won't work because `SafeExceptionReporterFilter` will not exist and raise a `NameError` instead of an `AttributeError`.

Why if the `ImportError` gives us all the information not just use that ?

```python
try:
    from django.views.debug import SafeExceptionReporterFilter
except ImportError:
    # Django < 3.0
    from django.views.debug import get_safe_settings
else:
    def get_safe_settings():
        return SafeExceptionReporterFilter().get_safe_settings()
```

Having the separate method on the class only makes sense to me if one would expect users to override that.. which I don't think will happen... then something like `settings.PRINT_SETTINGS_SAFE_FUNC` makes more sense so people have a way to override it... (but we also don't need that right now)

django_extensions/management/commands/print_settings.py
```diff
@@ -26,34 +31,47 @@ def add_arguments(self, parser):
...
+
+        show_secrets, output_format, indent = self.get_defaults(options)
+
         a_dict = {}
 
-        for attr in dir(settings):
-            if self.include_attr(attr, options['setting']):
-                value = getattr(settings, attr)
-                a_dict[attr] = value
+        if show_secrets:
```
trbs, line 60: We should be able to do this cleaner.... it smells like a lot of code duplication to me...

Makes more sense to do something like:

```python
if show_secrets:
    settings_dct = {k: getattr(settings, k) for k in dir(settings) if k.isupper()}
else:
    settings_dct = get_safe_settings()
```

trbs:
While looking at your pull request I did some cleanup in `print_settings` as that was needed and should make implementing your PR a lot cleaner.

Regretfully this does mean you will need to rebase....

(If you do not want to do this I can understand that and I can implement `hide secrets` directly on top of my most recent changes, but I would hate to see you lose attribution in git for your work !)

django_extensions/management/commands/print_settings.py
luchiago: Hmm, I liked your suggestion. I'll add it. Thanks for help me with this!

django_extensions/management/commands/print_settings.py
luchiago: Sure, I agree!

luchiago:
@trbs No problem at all! I'll rebase it

django_extensions/management/commands/print_settings.py
luchiago: @trbs Actually, I made a mistake.`Django < 3.0` does have the class `SafeExceptionReporterFilter`, but the method `get_safe_settings` is outside this class in those versions, only in recent versions the `SafeExceptionReporterFilter` does have the method `get_safe_settings`. Look at these links:
https://github.com/django/django/blob/stable/2.2.x/django/views/debug.py#L125
https://github.com/django/django/blob/stable/2.2.x/django/views/debug.py#L72

Even in Django == 3.0
https://github.com/django/django/blob/stable/3.0.x/django/views/debug.py#L72
https://github.com/django/django/blob/stable/3.0.x/django/views/debug.py#L125

Only in Django 3.1 occurs the change
https://github.com/django/django/blob/stable/3.1.x/django/views/debug.py#L108

That's why I used `AttributeError`

trbs: I see... maybe I made a typo when I got the `ImportError` before...
But we can still use the same style:

```python
try:
    from django.views.debug import SafeExceptionReporterFilter

    def get_safe_settings():
        return SafeExceptionReporterFilter().get_safe_settings()
except AttributeError:
    # Django < 3.1
    from django.views.debug import get_safe_settings
```

Alternatively we can add this to `django_extensions/compat.py` and keep this file a little bit cleaner :-)

tests/management/commands/test_print_settings.py
```diff
@@ -34,6 +35,16 @@ def test_with_multiple_setting_args(capsys):
...
 
+def test_with_not_show_secrets(capsys):
+    call_command(
+        'print_settings',
+        'SECRET_KEY',
+    )
+
+    out, err = capsys.readouterr()
+
+    assert '*******' in out
```
trbs, line 46: Can we make the check a little bit more explicit ?

eg use `--format=pprint` that would allow checking for `{'SECRET_KEY': 'dummy'}`

```
$ ./manage.py print_settings --format=pprint SECRET_KEY
{'SECRET_KEY': 'dummy'}
```

trbs: and adding an explicit check for the variant with `--show-secrets` would be even cooler ;-)

django_extensions/management/commands/print_settings.py
```diff
@@ -26,34 +31,47 @@ def add_arguments(self, parser):
...
             '--indent',
-            default=4,
+            default=DEFAULT_INDENT,
             dest='indent',
             type=int,
             help='Specifies indent level for JSON and YAML'
         )
+        parser.add_argument(
+            '--show-secrets',
+            default=DEFAULT_SECRETS,
```
trbs, line 47: use `action='store_true'` like the `--fail` option that way we don't need `DEFAULT_SECRETS` and the cli parser knows how to do the "right-thing"

luchiago: Right. Thanks for the tip!

tests/management/commands/test_print_settings.py
luchiago: Sure!

trbs:
Any interest on trying to pick this back up again to try and get it merged ?

luchiago:
> Any interest on trying to pick this back up again to try and get it merged ?

Hi @trbs. Sorry for the delayed response. It's been a while, so but I'll try to catch up and finally merge this PR

PR #1554, opened 2020-07-16
show_dsn: Fix crash with django-postgres-extra

tony:
https://django-postgres-extra.readthedocs.io/en/latest/

Manager doesn't follow the naming convention, but is still compatible with postgres DSN


trbs:
Thanks !

PR #1556, opened 2020-07-20
[mock] Use unitest mock module now in std lib

Djailla:
With current compatibility with Python >= 3.5, no need to use mock external module, it's integrated in the python standard library.


trbs:
Awesome thanks !

Issue #1467, opened 2020-01-20
--database option to be supported on the applicable functionalities


subramanyamVemu:
For most of the command line utilities it would be good to have the --database "<database>"

luchiago:
Hi. @subramanyamVemu @trbs I would like to tackle this issue, but I need more information. What should be the output of this command? Just show the database that is being used?

trbs:
> Hi. @subramanyamVemu @trbs I would like to tackle this issue, but I need more information. What should be the output of this command? Just show the database that is being used? And where this command should be available?

I am not sure I understand the question.

I think the point of this issue is twofold:
1) Support Django database routing for commands that do something with the database.
2) Make it uniform and work in the same way as Django does.

Currently `syncdata`, `sqlcreate`, `sqldsn`, `reset_schema`, `drop_tests_database` and `reset_db` have a `--database` option. (I use committed a patch to deprecate old style --router and consistently use `DEFAULT_DB_ALIAS`)

Other commands might still need a `--database` option and only work on the `DEFAULT_DB_ALIAS` for now.

luchiago:
@trbs Interesting. I sent a PR #1559 with a change on `print_settings.py`. I misunderstand the request and did another thing. Can we discuss the commands that need this option?

trbs:
Most likely any command using things like `from django.db import connection` or `from django.db import connections` and does not have a `--database` is a likely candidate. You can grep for it with something like: `$ ack 'connection' django_extensions/management/commands` (Replace `ack` with `grep` if you do not have `ack` installed)

A few possible candidates of the top of my head:
- `delete_squashed_migrations`
- `list_model_info`
- `sqldiff` (but also one of the most complicated pieces... not for the feint of heart to touch :-) )

luchiago:
@trbs Okay, I will work on this.

PR #1559, opened 2020-07-24
[feat] Database Option

luchiago:
Feature requested in the #1467. I added a new option in `print_settings.py` to show the current database. But we need to approve and merge the #1553 PR, because I made some refactoring in this file through this PR

### DIFF
--- a/django_extensions/management/commands/print_settings.py
+++ b/django_extensions/management/commands/print_settings.py
@@ -10,11 +10,11 @@ import json
 
 from django.conf import settings as default_settings
 from django.core.management.base import BaseCommand, CommandError
-
-from django_extensions.management.utils import signalcommand
 from django.views.debug import SafeExceptionReporterFilter
 
-DEFAULT_FORMAT = 'simple'
+from django_extensions.management.utils import signalcommand
+
+DEFAULT_FORMAT = "simple"
 DEFAULT_INDENT = 4
 DEFAULT_SECRETS = False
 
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/print_settings.py
```diff
@@ -26,34 +29,37 @@ def add_arguments(self, parser):
...
-            default=4,
+            default=DEFAULT_INDENT,
             dest='indent',
             type=int,
             help='Specifies indent level for JSON and YAML'
         )
+        parser.add_argument(
+            '--show-secrets',
+            default=DEFAULT_SECRETS,
+            dest='show secrets',
```
trbs: please do not use spaces, use underscores.

django_extensions/management/commands/print_settings.py
```diff
@@ -111,3 +111,13 @@ def include_attr(attr, settings):
     def print_simple(a_dict):
         for key, value in a_dict.items():
             print("%-40s = %r" % (key, value))
+
+    @staticmethod
+    def safe_settings():
+        try:
+            return SafeExceptionReporterFilter().get_safe_settings()
```
trbs, line 131: If django < 3.0 does not have this class then the import above will already have failed and this code is meaningless.

Either the comment below is wrong or we should catch `ImportError` during importing.

FYI: This entire code can be replaced to function at startup time which seem to be preferred to this way if it served no additional function.

django_extensions/management/commands/print_settings.py
```diff
@@ -47,11 +47,17 @@ def add_arguments(self, parser):
             type=bool,
             help="Specifies if should be revealed the value of secrets",
         )
+        parser.add_argument(
+            "--database",
+            dest="show database",
```
trbs, line 52: no spaces, just underscores :-)

trbs:
I guess I don't understand what purpose this PR serves... `DATABASES` is already included on the output of `print_settings` ?

luchiago:
> I guess I don't understand what purpose this PR serves... `DATABASES` is already included on the output of `print_settings` ?

@trbs As I said on #1467, I misunderstand the purpose. Let's close this PR and I can do the right one.

Issue #1081, opened 2017-07-27
Make runserver_plus survive syntax errors


maxnordlund:
Currently it crashes and you need to manually restart it, it would be nice if this wasn't the case. Not sure how to go about it though.

trbs:
We would welcome a pull-request to add this (preferable with a cli option for it)

maxnordlund:
Do you have any idea how to do it though?

twidi:
the django `runserver` command implements this since 2012 (https://github.com/django/django/commit/c2a6b2a43fb8f32cff078c823054e2e65f325464) so this should be possible here too, as we can expect `runserver_plus` to only add (the `_plus` part ;) ) features to the default django command, not to remove some :)

jrd:
Ahhh I always use `runserver_plus` so I didn't realized that `runserver` survives syntax error. I will check if I can propose a patch then 😉

skorokithakis:
@trbs can this be reopened until it is solved? It's currently my biggest source of frustration with django-extensions.

PR #1575, opened 2020-08-28
Make runserver_plus command survive syntax and configuration errors

Jenselme:
This will make it behave like the standard runserver command which doesn't
exit when we have a syntax or configuration error but simply wait for the
error to be fixed before the site is accessible again.

This should solve #1081

The solution turned out to be quite simple. Does it seem acceptable to you?

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -8,7 +8,7 @@ import sys
 import django
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management.base import BaseCommand, CommandError, SystemCheckError
 from django.core.servers.basehttp import get_internal_wsgi_application
 from django.dispatch import Signal
 from django.utils.autoreload import get_reloader
@@ -251,10 +251,13 @@ class Command(BaseCommand):
 
         if self.show_startup_messages:
             print("Performing system checks...\n")
-        if hasattr(self, 'check'):
-            self.check(display_num_errors=self.show_startup_messages)
-        else:
-            self.validate(display_num_errors=self.show_startup_messages)
+        try:
+            if hasattr(self, 'check'):
+                self.check(display_num_errors=self.show_startup_messages)
+            else:
+                self.validate(display_num_errors=self.show_startup_messages)
+        except (SyntaxError, SystemCheckError) as e:
+            self.stderr.write("SyntaxError occurred during system checks: " + str(e), ending="\n")
         try:
             self.check_migrations()
         except ImproperlyConfigured:
--- a/setup.py
+++ b/setup.py
@@ -155,7 +155,7 @@ setup(
     tests_require=[
         'Django',
         'Werkzeug',
-        'factory-boy',
+        'factory-boy<3.0.0',
         'mock',
         'pytest',
         'pytest-cov',


trbs:
Looks good, did you more or less copy this solution from the run server code ?

trbs:
Not sure if catching the exception and ignoring the exception is the right thing to do here....

I regretfully do not have the time right now but I would like to see how exactly `runserver` handles this. From first glance it does not seem to capture any `SystemCheckError` so don't think it should get passed ./core/management/commands/runserver.py lines 118-154/158... Think `runserver` just lets the thread exit but the reloader stays active 

I agree there are issues with this in `runserver_plus` currently... For example in my tests I could not see a simple exception message because of the clearing of line endings at `runserver_plus.py:128`... (sigh, so much to improve ;-) )

But I am not sure that ignoring the error and continuing is the best solution...

Jenselme:
> Not sure if catching the exception and ignoring the exception is the right thing to do here....

Neither am I. But I didn't see any other way to solve this.

> From first glance it does not seem to capture any SystemCheckError so don't think it should get passed ./core/management/commands/runserver.py lines 118-154/158... Think runserver just lets the thread exit but the reloader stays active

I didn't really look on this was done. I just witnessed the behavior and tried to reproduce it. It seemed like the most obvious way to change this.

From what I see [here](https://github.com/django/django/commit/c2a6b2a43fb8f32cff078c823054e2e65f325464#diff-46e69f287173eef41fcbfeba05501954R87), I should catch a bit more errors to correctly match `runserver`. From what I see, once the error is catch, they just add the problematic files to a list so the watcher correctly watches them. Since `werkzeug` seems to be doing this just fine, I don't see what we can do besides ignoring it.

trbs:
For the sake of argument, say I wrote a check: ( https://docs.djangoproject.com/en/3.1/topics/checks/ )

```python
@register()
def make_very_sure_runserver_does_not_use_production_database(app_configs, **kwargs):
    ...
```

Now it fails... displays the warning but continuous running `runserver_plus` and potentially doing nasty things to our production systems ;-)

trbs:
Could you please try https://github.com/django-extensions/django-extensions/commit/ae42e864438b2ec42202f608240ceb3bc4e57f58

Does this solve the issue for you ?

Jenselme:
The command doesn't crash but once we have a syntax error, even if I fix it, I have to kill the command and restart it for the fix to be taken into account. So it isn't really better than before.

BTW, I noticed that the standard `runserver` kills the server on syntax error and wait for the error to be fixed before to restart the server. I don't know whether it helps.

trbs:
I can reproduce the issue with `SyntaxError`... not entirely sure how to fix it quickly :-)

For now I removed the check for `SyntaxError` so it only catches `SystemCheckError`.
Think it's better to stop `runserver_plus` then showing an error which does not auto reload....

trbs:
Think for now we can close the PR... 
If you might find a way to also support `SyntaxError` or other exceptions please submit another PR ! 

Thanks for all the help on this and making Django Extensions better !

Jenselme:
No idea right now. I'll think I'll just rely on bash in the short term to relaunch `runserver_plus` when need. What I find strange is that with my patch it worked. Since both are very close, I don't understand why. On syntax error, I never got a message about the code being reloaded though :thinking: 

Thanks anyway!

trbs:
In https://github.com/pallets/werkzeug/blob/7f6088d7f774bd6e7d3f6cede0d7a284ba5f734c/src/werkzeug/_reloader.py#L13 `sys.modules` is used, I bet after the `SyntaxError` that that module is not on the list ;-)

A solution might be to just pass the file for the module that yielded the syntax error to `extra_files` in `run_simple`. However I guess it will still not trigger reload on other files that where still not loaded since the `SyntaxError` interrupted the process.

Jenselme:
Since the server can't start with the syntax error in that file, is it a big issue? That should be testable too. I'll try to test this tomorrow to see how it goes.

Jenselme:
Based on your last patch, this seems to work perfectly:
```patch
diff --git i/django_extensions/management/commands/runserver_plus.py w/django_extensions/management/commands/runserver_plus.py
index 560d13bc..c205c45b 100644
--- i/django_extensions/management/commands/runserver_plus.py
+++ w/django_extensions/management/commands/runserver_plus.py
@@ -4,6 +4,7 @@ import os
 import re
 import socket
 import sys
+import traceback
 
 import django
@@ -276,9 +277,21 @@ class Command(BaseCommand):
         if self.show_startup_messages:
             print("Performing system checks...\n")
 
+        files_in_error = set()
         try:
             self.check(display_num_errors=self.show_startup_messages)
-        except SystemCheckError as exc:
+        except (ImportError, IndentationError, SyntaxError, NameError,
+                TypeError, AttributeError, SystemCheckError) as exc:
+            # Inspired by https://github.com/django/django/commit/c2a6b2a43fb8f32cff078c823054e2e65f325464#diff-46e69f287173eef41fcbfeba05501954R89  # noqa: E501
+            et, ev, tb = sys.exc_info()
+            if getattr(ev, 'filename', None) is None:
+                # get the filename from the last item in the stack
+                filename = traceback.extract_tb(tb)[-1][0]
+            else:
+                filename = ev.filename
+            if filename not in files_in_error:
+                files_in_error.add(filename)
+
             self.stderr.write("SystemCheckError occurred during system checks: " + str(exc), ending="\n\n")
             handler = self.get_systemcheckerror_handler(str(exc), error_class=type(exc), **options)
         else:
@@ -364,7 +377,7 @@ class Command(BaseCommand):
             handler,
             use_reloader=use_reloader,
             use_debugger=True,
-            extra_files=self.extra_files,
+            extra_files=self.extra_files.union(files_in_error),
             reloader_interval=reloader_interval,
             reloader_type=reloader_type,
             threaded=threaded,
```

Even if I modify files other than the one with the error, the project is correctly reloaded.

trbs:
Taking inspiration from your last comment I tried to get even closer to `runserver` :-)

Could you please check with latest master ( https://github.com/django-extensions/django-extensions/commit/6fdddd39b12674f60b8f92191a7b68c8a55a7ae5 ) if this works for you ?

Jenselme:
@trbs it seems to work perfectly. Do you plan to make a new release with it soon?

trbs:
FYI: New release should be out shortly :-)

PR #1579, opened 2020-09-05
add uuid model & soft delete model & manager

thomas545:


trbs:
Thanks !

I like the idea (specially for `UuidModel`) however I am thinking that this normally is part of a larger effort for most applications...

thomas545:
> Thanks !
> 
> I like the idea (specially for `UuidModel`) however I am thinking that this normally is part of a larger effort for most applications... Maybe just having the abstract model is not helping people enough ? might benefit more from some more stuff around it like larger documentation and shortcuts to help using uuid's instead of pk's...

Thank You 
but I'm just thinking to use uuid with pk isn't instead of pk.

camilonova:
Django has https://docs.djangoproject.com/en/4.0/ref/models/fields/#uuidfield

I'm closing this. Thank you.

PR #1583, opened 2020-09-11
Improve sqlcreate to supports PostgreSQL unix domain socket

luzfcb:
### The problem:

Currently the `sqlcreate` generate invalid SQL statements, like:

```sql
CREATE USER  WITH ENCRYPTED PASSWORD '' CREATEDB;
CREATE DATABASE mydatabase WITH ENCODING 'UTF-8' OWNER "";
GRANT ALL PRIVILEGES ON DATABASE mydatabase TO ;
```

if a Django database is configured to use Unix domain socket, like:

```python
DATABASES = {
    'default': {
        'ENGINE': 'django.db.backends.postgresql',
        'NAME': 'mydatabase',
        'USER': '',
        'PASSWORD': '',
        'HOST': '',
        'PORT': '',
    }
}
```
then errors can be obtained when using `python manage.py sqlcreate |  sudo -u postgres psql -U postgres` like:

```sql
ERROR:  syntax error at or near "WITH"
LINE 1: CREATE USER  WITH ENCRYPTED PASSWORD '' CREATEDB;
                     ^
ERROR:  zero-length delimited identifier at or near """"
LINE 1: CREATE DATABASE mydatabase WITH ENCODING 'UTF-8' OWNER "";
                                                             ^
ERROR:  syntax error at or near ";"
LINE 1: GRANT ALL PRIVILEGES ON DATABASE mydatabase TO ;
```

### Solution

The solution I propose with this pull-request, is to change the logic so that, if `USER` or `PASSWORD` is not informed on django database configuration, then only SQL `CREATE DATABASE` will be generated, like:

`python manage.py sqlcreate`
```sql
-- Assuming that unix domain socket connection mode is being used because
-- USER or PASSWORD are blank in Django DATABASES configuration.
CREATE DATABASE mydatabase WITH ENCODING 'UTF-8';
```

--
Pull-request sponsored by 
[image]( https://www.feldroy.com/)

### DIFF
--- a/django_extensions/management/commands/sqlcreate.py
+++ b/django_extensions/management/commands/sqlcreate.py
@@ -70,11 +70,20 @@ The envisioned use case is something like this:
         elif engine in ('postgresql', 'postgresql_psycopg2', 'postgis'):
             if options['drop']:
                 print("DROP DATABASE IF EXISTS %s;" % (dbname,))
-                print("DROP USER IF EXISTS %s;" % (dbuser,))
+                if dbuser:
+                    print("DROP USER IF EXISTS %s;" % (dbuser,))
 
-            print("CREATE USER %s WITH ENCRYPTED PASSWORD '%s' CREATEDB;" % (dbuser, dbpass))
-            print("CREATE DATABASE %s WITH ENCODING 'UTF-8' OWNER \"%s\";" % (dbname, dbuser))
-            print("GRANT ALL PRIVILEGES ON DATABASE %s TO %s;" % (dbname, dbuser))
+            if dbuser and dbpass:
+                print("CREATE USER %s WITH ENCRYPTED PASSWORD '%s' CREATEDB;" % (dbuser, dbpass))
+                print("CREATE DATABASE %s WITH ENCODING 'UTF-8' OWNER \"%s\";" % (dbname, dbuser))
+                print("GRANT ALL PRIVILEGES ON DATABASE %s TO %s;" % (dbname, dbuser))
+            else:
+                print(
+                    "-- Assuming that unix domain socket connection mode is being used because\n"
+                    "-- USER or PASSWORD are blank in Django DATABASES configuration."
+                )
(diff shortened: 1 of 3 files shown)


luzfcb:
@trbs Since this changes the current default behavior (which is to generate invalid SQL), perhaps this change should be classified as breaking change.

django_extensions/management/commands/sqlcreate.py
```diff
@@ -70,11 +70,20 @@ def handle(self, *args, **options):
...
+                    print("DROP USER IF EXISTS %s;" % (dbuser,))
+
+            if dbuser and dbpass:
+                print("CREATE USER %s WITH ENCRYPTED PASSWORD '%s' CREATEDB;" % (dbuser, dbpass))
+                print("CREATE DATABASE %s WITH ENCODING 'UTF-8' OWNER \"%s\";" % (dbname, dbuser))
+                print("GRANT ALL PRIVILEGES ON DATABASE %s TO %s;" % (dbname, dbuser))
+            else:
+                print(
+                    "-- Assumes that unix domain socket connection mode is being used because "
+                    "USER or PASSWORD is blank in django DATABASES configuration."
```
trbs, line 83: Couple of small suggestions / nitpicks:

- Better to talk with authority and say "-- **Assuming** that"
- Think that should be "USER or PASSWORD **are** blank" ?
- Django with capital D "blank in **D**jango DATABASES"

luzfcb: Thanks for the review. Fixed.

luzfcb:
@trbs The failing tests are not related to the changes included on this pull-request.

trbs:
tests should be fixed now :-)

trbs:
Yeah I agree on the breaking change thing. but since this was already broken I don't really mind ;-)

luzfcb:
@trbs Thanks for all your hard work to maintain this library.

PR #1588, opened 2020-09-29
Pipchecker retry

### LINKED ISSUE #1587: pipchecker - PyPI rate limit
django-extensions==3.0.9
pip==20.2.3

NB: I'm fairly certain this is *not* related to existing GitHub-specific rate-limiting issues.

Pointing pipchecker at a sufficiently long `requirements.txt` results in the following error (redacted for readability):

```
  File "~/.virtualenvs/env/lib/python3.7/site-packages/django_extensions/management/commands/pipchecker.py", line 111, in handle
    self.check_pypi()
  File "~/.virtualenvs/env/lib/python3.7/site-packages/django_extensions/management/commands/pipchecker.py", line 148, in check_pypi
    available = pypi.package_releases(req["pip_req"].name, True) or pypi.package_releases(req["pip_req"].name.replace('-', '_'), True)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 1112, in __call__
    return self.__send(self.__name, args)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 1452, in __request
    verbose=self.__verbose
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 1154, in request
    return self.single_request(host, handler, request_body, verbose)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 1170, in single_request
    return self.parse_response(resp)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 1342, in parse_response
    return u.close()
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/xmlrpc/client.py", line 656, in close
    raise Fault(**self._stack[0])
xmlrpc.client.Fault: <Fault -32500: 'HTTPTooManyRequests: The action could not be performed because there were too many requests by the client. Limit may reset in 58 seconds.'>
```

I believe this is related to recently added rate limiting on PyPI (can't find a news announcement but swear I remember seeing something).

In terms of a PR, would a general approach of "back off and retry" be appropriate/welcome? (Unclear if the response actually includes the retry time in e.g. a header)

mplumador:
Fixes #1587 

Unfortunately the response is an [XML-RPC Fault](https://docs.python.org/3/library/xmlrpc.client.html#xmlrpc.client.Fault) object, which does not have headers for retrieving the time. Pypi seems to have a 60 second retry limit which made an exponential backoff approach inefficient since it will not process the requests any sooner than 60 seconds. Also it appears that the XML-RPC api for pypi is [up for depreciation ](https://warehouse.pypa.io/api-reference/xml-rpc/#pypi-s-xml-rpc-methods) which may require a more significant PR in the future. 

I'm a little uncertain about how to test something like this, but I have included a new test that writes a requirements.txt file that I found reproduced the issue.


trbs:
One can only hope that we can stop using XML-RPC as soon as possible 👍 
As you say this is a larger efforts and we need to find a sponsor for it... (time wise or financial)
Otherwise it will be likely that we will remove `pipchecker` when it breaks as XML-RPC gets removed someday in the future.

For now the PR seems fine for me... however I am thinking that we should print some message to the user... as 60 seconds is a long time to wait without anything happening :-(

trbs:
Thanks !

picturedots:
Thanks @mplumador  for this update -- it's working again now for me, but very slow (~20 minutes for a 20 line file, on account of the 60 second retry).

I've been using [`pip list --outdated`](https://pip.pypa.io/en/stable/reference/pip_list/) as a workaround which runs instantly, but does not take a requirements file as an input.

PR #1590, opened 2020-10-09
Add django-zero-downtime-migrations Postgres DB engine

malthejorgensen:
From the repo: 
https://github.com/tbicr/django-pg-zero-downtime-migrations

I got an error when running the `reset_db` command while using this engine, which is why I'd like to add it to the list :)


malthejorgensen:
This may be better solved by simply adding more docs for the `reset_db`-command as in the PR over here:
https://github.com/django-extensions/django-extensions/pull/1591

I'll leave that up to the maintainers :)

PR #1591, opened 2020-10-09
Add more documentation for `reset_db`-command

malthejorgensen:
Adds full page to document the `reset_db`-command. It does not cover most options, but that follows the pattern from other pages.

For me, the big thing is documenting the `DJANGO_EXTENSIONS_RESET_DB_*_ENGINES`-settings which is added to this page.


trbs:
Thanks !

I actually have a TODO in the back of my mine to improve this engine detection by using something like `inspect.getmro(...)`. This should allow any engine based on `django.db.backends.` to work without to much of a hassle with these `*_ENGINES` lists in settings...

PR #1601, opened 2020-11-10
Adding code to run on Powersystem

genisysram:
Hi
Here is my contribution to your code, its working good on powersystems.

Thanks for the code, its working good.
## What do these changes do?
Added Architecture "ppc64le"
## Are there changes in behavior for the user?
No

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -228,18 +228,33 @@ jobs:
       arch: ppc64le
    
   allow_failures:
-    - python: 3.6
-      env: TOXENV=py36-djmaster
-    - python: 3.7
-      env: TOXENV=py37-djmaster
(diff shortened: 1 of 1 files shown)


trbs:
Pretty cool ;-)

Seems couple of tests are failing but I have a larger issues.... not specifically with your PR but with Travis...

I plan to migrate from Travis to something else like Github Actions.
Since Travis is unusable nowadays...

Not sure if Github Actions (or something else which is Free) supports PPC and ARM architectures... if so I would be very happy to add those there...

trbs:
To answer my own question... no Github Action does not support anything; https://github.community/t/testing-against-multiple-architectures/17111/2

PR #1605, opened 2020-11-18
fix issue with RUNSERVER_PLUS_EXTRA_FILES setting parsing

omaraltayyan:
django's setttings don't support subscript access which causes an exception in django 3.1 if this setting is used


trbs (approved):

trbs:
Thanks !

Issue #1577, opened 2020-09-04
specify app_names in GRAPH_MODELS settings


sdementen:
is it possible to specify the app_names in the GRAPH_MODELS setting to allow to simply run "manage graph_models" ?

trbs:
Sorry, this is not possible at the moment. 

Happy to accept a PR for this change though :-)

PR #1610, opened 2020-12-07
proposal for issue #1577

sdementen:
To allow to have default app_labels in the GRAPH_MODELS:
- added option --app-labels accepting one or more arguments
- get default for app-labels from GRAPH_MODELS["app-labels"]
- support both --app-labels and positional arguments for app_label

### DIFF
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -44,16 +44,15 @@ class Command(BaseCommand):
         --disable-fields can be set in settings.GRAPH_MODELS['disable_fields'].
         """
 
-        class EmptyList(list):
+        class DefaultList(list):
             def extend(self, __iterable):
                 """On first extend, replace the list by the __iterable (if not []).
                 Otherwise, behave like a normal list."""
-                _first_extend = getattr(self, "_first_extend", False)
-                if not _first_extend and __iterable:
-                    self[:] = __iterable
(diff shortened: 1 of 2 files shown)


trbs:
Could you make this simpler ?
Thinking that some thing like this would be enough:
```diff
diff --git a/django_extensions/management/commands/graph_models.py b/django_extensions/management/commands/graph_models.py
index 8cb4a57..6da4a08 100644
--- a/django_extensions/management/commands/graph_models.py
+++ b/django_extensions/management/commands/graph_models.py
@@ -195,7 +195,11 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         args = options['app_label']
         if not args and not options['all_applications']:
-            raise CommandError("need one or more arguments for appname")
+            default_app_labels = getattr(settings, 'GRAPH_MODELS', {}).get("APP_LABELS")
+            if default_app_labels:
+                args = default_app_labels
+            else:
+                raise CommandError("need one or more arguments for appname")

         # Determine output format based on options, file extension, and library
         # availability.
```

Also we need to update the docs as well..

sdementen:
you're right, much simpler!
what do you want me to do? another PR ? or you convert your patch in a PR ?

sdementen:
is this better ? I used app_labels instead of APP_LABELS to be consistent with the casing of the other options

trbs:
Thanks !

PR #1612, opened 2020-12-18
DOC: Make notion of deprecated passwd in docs.

hwalinga:
`passwd` is deprecated, but it is not specified as such in the docs.

We can also just remove the entry.

We can also add a link to official documentation of Django `changepassword`.


trbs:
Thanks !

PR #1621, opened 2021-01-27
Fix to keep the compatibility with Python 2

AndreAngelucci:


ulgens:
The project doesn't support old Python for a while and I'm not sure trying to get that support back will end with the satisfying result for anyone 🙂

AndreAngelucci:
Got it, thanks for the feedback!

PR #1624, opened 2021-02-01
Remove redundant markers

ulgens:
There were several test skip checks aiming towards to python and Django versions which are not supported by this project anymore. I removed those checks.


trbs (approved):

trbs:
Thanks !

PR #1628, opened 2021-02-12
fix shell_plus --command globals / locals error

terencehonles:
When running a "more complex" command via `--command` shell_plus has historically passed an empty globals and the imports as locals. This will cause Python to act strangely for certain inputs such as list comprehensions. This was fixed in Django's shell command [1] and by passing the imports as the only arg Python will use them as both the locals and the globals which is as if the `exec` was happening at the module level.

[1]: https://code.djangoproject.com/ticket/32183

See also: https://stackoverflow.com/questions/45132645/list-comprehension-in-exec-with-empty-locals-nameerror


terencehonles:
@trbs not sure if you have a chance to review / merge this one too.

trbs:
Thanks !

PR #1632, opened 2021-02-18
update shell_plus to not save ipython history when using Jupyter

terencehonles:
xref: https://github.com/jupyter/notebook/issues/3001#issuecomment-429760689

### DIFF
--- a/django_extensions/management/commands/shell_plus.py
+++ b/django_extensions/management/commands/shell_plus.py
@@ -204,7 +204,7 @@ class Command(BaseCommand):
 
         return {'django_extensions': ks}
 
-    def run_notebookapp(self, app_init, options, use_kernel_specs=True):
+    def run_notebookapp(self, app_init, options, use_kernel_specs=True, history=True):
         no_browser = options['no_browser']
 
         if self.extra_args:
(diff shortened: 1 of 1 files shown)


django_extensions/management/commands/shell_plus.py
```diff
@@ -235,6 +235,10 @@ def run_notebookapp(self, app_init, options, use_kernel_specs=True):
         if not use_kernel_specs:
             notebook_arguments.extend(ipython_arguments)
 
+        # disable history if not already configured in some other way
+        if not history and not any(i.startswith('--HistoryManager') for i in ipython_arguments):
```
trbs, line 239: small nitpick.. could we use 'e' instead of 'i' here ? (as in `for e in ipython_arguments`)

as short var names like `i`, `j`, `n`, `m`, etc indicate something numeric in my mind and things like `e` or `o` would be sort for 'element' or 'object'... (or maybe 'option' in this case...)

otherwise a bit longer but clearer `for arg in ipython_arguments` would also be a possibility ;-)

terencehonles: will do

trbs:
Thanks !

PR #1633, opened 2021-02-24
Set minimum django version

mlissner:
This adds django 2.2 to the requirements list. Without it listed like this, dependency management tools will think that this versions > 3.0 of django extensions can be installed on older versions of django. In our case, dependabot recommended that we upgrade to version 3.11 of this library, but we do not have a compatible version of django (yes, I know, we should; we're getting there!)

It'd be great to get this small fix in place so that dependabot and other compatibility checkers can ensure that only compatible versions are installed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ setup(
     cmdclass=cmdclasses,
     package_data=package_data,
     python_requires=">=3.6",
-    install_requires=[],
+    install_requires=["Django>=2.2"],
     extras_require={},
     classifiers=[
         'Development Status :: 5 - Production/Stable',


trbs:
Thanks for the PR and sorry for the delay.

There where some reasons in the past why I didn't add it but I cannot find a good reason not to add the minimum requirement today ;-)

New release should be out (relatively) soon.

However this would only solve this issues in the future... let me try to create a 3.1.1post1 as well...

trbs:
https://pypi.org/project/django-extensions/3.1.1.post1/

Hope this works for dependabot (etc), the post1 release should indicate to them they should use the new package metadata... https://pythonpackaging.info/07-Package-Release.html

PR #1637, opened 2021-03-22
Have drop_test_database delete all cloned test databases

### LINKED ISSUE #1636: The drop_test_database management command does not drop all test databases when using --parallel flag
On a system with multiple cores, running `manage.py test --parallel --keepdb`, produces multiple test databases (one for each core):

    test_<dbname>
    test_<dbname>_1
    test_<dbname>_2
    ...etc

After running `manage.py drop_test_database`, only `test_<dbname>` gets dropped and the other test databases remain intact. This could cause a state where future test suite runs are running parts of the test suite against different versions of the database.

Would it be reasonable to update the command to drop all test databases that follow the expected naming convention, or alternatively if it's determined the command is not intended to work well with the `--parallel` flag, to document this caveat?

abbottc:
When using the --parallel flag on 'manage.py test', multiple test
databases are created. Previously, only the primary test database was
removed. This commit also removes all additional cloned databases if
they exist.

Fixes #1636

### DIFF
--- a/django_extensions/management/commands/drop_test_database.py
+++ b/django_extensions/management/commands/drop_test_database.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+from itertools import count
 import os
 import logging
 import warnings
@@ -82,12 +83,15 @@ class Command(BaseCommand):
         verbosity = options["verbosity"]
         if options['interactive']:
             confirm = input("""
-You have requested to drop the test database.
+You have requested to drop all test databases.
 This will IRREVERSIBLY DESTROY
-ALL data in the database "%s".
+ALL data in the database "{db_name}"
+and all cloned test databases generated via
+the "--parallel" flag (these are sequentially
+named "{db_name}_1", "{db_name}_2", etc.).
 Are you sure you want to do this?
 
-Type 'yes' to continue, or 'no' to cancel: """ % (database_name,))
+Type 'yes' to continue, or 'no' to cancel: """.format(db_name=database_name))
         else:
             confirm = 'yes'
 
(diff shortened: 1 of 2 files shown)


django_extensions/management/commands/drop_test_database.py
```diff
@@ -82,26 +83,61 @@ def handle(self, *args, **options):
...
+                # Reference: https://code.djangoproject.com/ticket/32582
+                return '{}_{}.{}'.format(filename, number, ext)
+
             try:
-                logging.info("Unlinking %s database" % engine)
-                if os.path.isfile(database_name):
-                    os.unlink(database_name)
+                for db_name in get_database_names(format_filename):
+                    if not os.path.isfile(db_name):
+                        break
```
trbs, line 135: Would `break` or `continue` be more appropriate here.... ?

(eg: say I deleted `foo_bar_test_2.db` myself but not `foo_bar_test_3.db`...)

trbs:
+1 on this PR.
Love that you updated the tests as well ! Thank you !

I will try to see if I can make some time to move `django-extensions` to Github Actions instead of Travis and get everything to work properly again.

django_extensions/management/commands/drop_test_database.py
abbottc: The issue is that a developer can supply any value to `--parallel`, and Django will create that many test databases up to a maximum of the number of TestCase classes that exist. I could create 1000 databases with `manage.py --parallel 1000 --keepdb`. The `drop_test_database` code has no knowledge of the value last used with `--parallel`, so some other strategy must exist to halt the loop that looks for test databases to delete. 

The most reasonable strategy to my mind is checking sequential existence, since the only guarantee we can rely on from Django's process is that the test database names are sequentially numbered. That does leave the caveat you point out that a developer could manually delete some test database in the middle and then run `drop_test_database`, but to me it seems that would be the developer's responsibility to clean up, since it's an odd thing to do.

Alternatively, an arbitrarily high upper limit on the possible number could be hard-coded into the loop, and the loop could always run that number of times, checking all database names. This approach to me is inferior, since it relies on guessing some value assumed to be higher than any developer would use, and is slower since it would need to run that number of database queries (or file existence checks in the case of sqlite) every time the command is run. And, despite the downsides, it still leaves its own caveat since some developer could always create a larger number of test database than the maximum we set on this loop and those won't get deleted.

If you do want the second strategy, let me know what value cap you want to use, and I'll implement that. Or if you have an idea for a different strategy to solve this problem, let me know. But my current recommendation would be what's implemented, and perhaps if you prefer, I could add something to the docs warning that the command may not remove everything if a developer manually deletes any test databases first.

abbottc:
Thanks for the kudos about testing. I'm big on that. And thanks for looking at the CI issues. Your work on maintaining a large open-source project is very highly appreciated.

trbs:
Would you mind rebasing master to see if GH Actions work properly for the PR ?

abbottc:
Hey, that looks great. All the CI gremlins are gone.

trbs:
Awesome.

Many thanks for the PR, updated test and detailed explanations ! :-)

trbs:
I will try to do a new release soon as well...
(If that doesn't happen in the next week or so feel free to nudge me on that ;-) )

abbottc:
And thanks for reviewing, merging, and handling the CI issue all in a timely manner. Very glad to have helped!

trbs:
https://pypi.org/project/django-extensions/3.1.2/

PR #1651, opened 2021-04-14
Fix sqldiff for missing field/index in model case

olegpesok:
When both the field and index are missing from the model, the function
would error out. Instead report as an index that should be removed.


trbs:
Thanks !

PR #1653, opened 2021-04-16
Fixed default_app_config deprecation

terencehonles:
This makes it easier for consumers that turn on all warnings to not have to filter this warning out.


trbs:
I just merged https://github.com/django-extensions/django-extensions/commit/a85bcac8687d52a8c2f47e3ff77b03525d7307d7

Think we can close this PR :-)

Thank you for the PR !

PR #1656, opened 2021-04-20
fix: Update codebase to use newer django syntax

abhiabhi94:
- all commands of django_extensions now inherit from a base command `_BaseDjangoExtensionsCommand`.
- this helps in defining constant at a single place.(The `_` is prefixed intentionally to indicate the `class` as an internal API.)
- handle django41 deprecation warnings.
- update some test code to use upgraded version of django code.

### DIFF
--- a/django_extensions/management/commands/clean_pyc.py
+++ b/django_extensions/management/commands/clean_pyc.py
@@ -4,15 +4,16 @@ import os
 from os.path import join as _j
 
 from django.conf import settings
-from django.core.management.base import BaseCommand, CommandError
+from django.core.management import CommandError
 
+from django_extensions.management import _BaseDjangoExtensionsCommand
 from django_extensions.management.utils import signalcommand
 
 
-class Command(BaseCommand):
+class Command(_BaseDjangoExtensionsCommand):
     help = "Removes all python bytecode compiled files from the project."
 
-    requires_system_checks = False
+    requires_system_checks = []  # type: ignore
 
     def add_arguments(self, parser):
         parser.add_argument(
--- a/tests/testapp/urls.py
+++ b/tests/testapp/urls.py
@@ -16,15 +16,15 @@ Including another URLconf
     1. Import the include() function: from django.conf.urls import url, include
     2. Add a URL to urlpatterns:  url(r'^blog/', include('blog.urls'))
 """
-from django.conf.urls import url
 from django.contrib import admin
 from django.contrib.auth import views as auth_views
+from django.urls import path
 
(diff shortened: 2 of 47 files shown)


tests/testapp/urls.py
```diff
@@ -16,15 +16,15 @@
...
+from django.urls import re_path
 
 login_view = auth_views.LoginView.as_view() if hasattr(auth_views, 'LoginView') else auth_views.login
 logout_view = auth_views.LogoutView.as_view() if hasattr(auth_views, 'LogoutView') else auth_views.logout
 
 urlpatterns = [
-    url(r'^login/$', login_view, {'template_name': 'login.html'}, name="login"),
-    url(r'^logout/$', logout_view, name="logout"),
-    url(r'^admin/', admin.site.urls),
+    re_path(r'^login/$', login_view, {'template_name': 'login.html'}, name="login"),
```
ulgens, line 27: Why not just `path`?

django_extensions/management/commands/clean_pyc.py
```diff
@@ -12,7 +12,7 @@
 class Command(BaseCommand):
     help = "Removes all python bytecode compiled files from the project."
 
-    requires_system_checks = False
+    requires_system_checks = []
```
abhiabhi94, line 15: in newer version of django, this value should be list/tuple of `Tags` or `__all__`. You can check the documentation [here](https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.BaseCommand.requires_system_checks).

For `mypy` linting to work well with backward comptability, I think one of the solution would be to do something of the form:
```python

from typing import List, Tuple, Union

from django.core.checks import Tags

requires_system_checks: Union[bool, List[Tags], Tuple[Tags], str]
```

Instead of making this change everywhere in the codebase, I think it would be better to create some sort of `BaseCommand`and let all commands inherit from it. 

Would it be acceptable? Any suggestions are welcome.

abhiabhi94:
I have had to make some changes to workaround `mypy` which might make the `diff` a bit more than it should have.

As of now, it seems that `mypy` doesn't necessarily work well with inherited class attributes, hence had to ignore it at certain places as a workaround.

```python
class A:
    a: List[]

class B:
    a = []  # <- mypy tends to complain about this
```

I have created there as well https://github.com/python/mypy/issues/10375, so for the time being, I think we can use the `type: ignore` thing as a workaround.

camilonova:
Looks like a big change and the codebase has changed a lot since it was first proposed. I'm closing this. Thank you.

PR #1660, opened 2021-04-24
ADD set of auto encrypting fields and Mixin to construct custom one

kamil1marczak:


camilonova:
I don't see this as a generic use case for this library. I'm closing it. Thank you.

PR #1661, opened 2021-04-25
Fix name of mixin in docs

ryanwilsonperkin:
Docs were referencing a class called UserPermissionMixin but it is actually called ModelUserFieldPermissionMixin


trbs:
Thanks !

PR #1662, opened 2021-04-27
Fix django.contrib.sites detection in set_default_site command when using AppConfig

azaghal:
The purpose of this pull request is to improve `django.contrib.sites` application detection in the `set_default_site` command. Currently the command depends on having the name hardcoded in `settings.INSTALLED_APPS`, which could lead to not being able to detect the application if it is installed through the AppConfig class/mechanism.

Instead of the simple string check, the changed implementation uses application registry.

While not something that is common (I literally found one project that listed the `django.contrib.sites.apps.SitesConfig` in its example config), the small fix I propose is probably more robust than the existing one.

### DIFF
--- a/django_extensions/management/commands/set_default_site.py
+++ b/django_extensions/management/commands/set_default_site.py
@@ -3,6 +3,7 @@ import socket
 
 from django.conf import settings
 from django.core.management.base import BaseCommand, CommandError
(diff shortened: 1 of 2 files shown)


trbs (approved):

trbs:
Thanks !

PR #1665, opened 2021-05-10
Add --relation-fields-only flag to graph_models

dankgen-tobias:
When using the management command `graph_models` for a database with a lot of models and a lot of fields, the resulting graph can get quite big and confusing. By introducing a flag `--relation-fields-only` only fields of the type `[ForeignKey, ManyToManyField, OneToOneField, RelatedField, OneToOneRel, ManyToOneRel]` are displayed. This leads to a more clean graph with a better overview that highlights the connections between models while hiding all other fields.

Since this is my first contribution to a public project ever I hope I contributed in the right way. Let me know if something should be improved or changed.


OmenApps:
A quick glance at the code - this looks great! I have 200+ models in one of my projects, so I understand exactly where you're coming from with this.

And you included tests, too!

I think the maintainers are backlogged a bit right now, but this looks like a pretty solid commit once they get to it.

BergiBerghold:
Looks good, I'd like to see this in django-extensions

PR #1675, opened 2021-07-30
feat: Add command to list latest migrations

torazem:
I've created this to make it easier for me to roll back to a particular state, e.g.:
```shell
python manage.py list_latest_migrations.py > rollback.txt
awk '{print "python manage.py migrate " $0}' rollback.txt
```
which would get me the list of commands to get me back to the current state.

If there's an easier way to do this, please let me know~

I am struggling to figure out how to write tests for this though.


camilonova:
No tests, No docs, No merge :)

PR #1690, opened 2021-11-25
Accept both --no-input and --noinput

PeterJCLaw:
This makes it easier for users who may be used to one spelling or the other and brings these commands compatibility with Django built-in commands (which all accept either spelling).


PeterJCLaw:
@camilonova apologies if you're not the right person to ping, is there anything I need to do to get this merged? I'm not sure what the PR process is for this repo.

camilonova:
Thank you!

PR #1693, opened 2021-12-14
Graph models options to get ERD with database data

fbuccioni:
A client need physical diagram of the database but I cannot generate one, so I took the plugin and add some options, if I don't follow the code lines please tell me about how can I do it, enlight my path :).


fbuccioni:
Updated code to main source code,

PR #1704, opened 2022-02-03
Fix one code format in documentation

rixx:
Currently, the `--` renders as – in the docs, which had me confused until I read the example more carefully.


trbs:
Thanks !

PR #1726, opened 2022-05-29
Add support for meta indexes and constraints in sqldiff

noamkush:
As title says, this adds support to sqldiff for models that specify indexes/unique constraints in new style:
``` python
class Meta:
    indexes = [models.Index(...)]
    constraints = [models.UniqueConstraint(...)]
```


trbs:
Thanks !

PR #1732, opened 2022-06-12
Sync Django trove classifiers with test setup

ulgens:
Completes missing parts from https://github.com/django-extensions/django-extensions/pull/1730/


camilonova:
Thank you

Issue #1099, opened 2017-09-15
AutoSlugField can't handle allow_unicode=True.


allieus:
AutoSlugField uses `django.template.defaultfilters.slugify` internally. This function can not handle the allow_unicode option.
https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/fields/__init__.py#L25

You must use `django.utils.text.slugify` instead of `django.template.defaultfilters.slugify`.

You must also specify the allow_unicode option in the following code.
https://github.com/django-extensions/django-extensions/blob/master/django_extensions/db/fields/__init__.py#L134

Thanks.

shukryzablah:
The documentation links to `django.utils.text.slugify` intead of `django.template.defaultfilters.slugify`

PR #1734, opened 2022-06-19
Update django slugify url

andyklimczak:
Incorrect url noted in https://github.com/django-extensions/django-extensions/issues/1099#issuecomment-867069345=


ulgens:
It's not the incorrect URL. One of them is the function and the other one is the template tag. The doc refers to the function.

PR #1739, opened 2022-07-08
Consider unique constraint expressions to get unique value

whitespy:
Currently, fields like AutoSlugField does not consider introduced in Django 4.0 UniqueConstraint.espressions to generate unique values. This PR resolves this issue.

### DIFF
--- a/django_extensions/db/fields/__init__.py
+++ b/django_extensions/db/fields/__init__.py
@@ -20,9 +20,11 @@ try:
 except ImportError:
     HAS_SHORT_UUID = False
 
+import django
+
 from django.conf import settings
 from django.core.exceptions import ImproperlyConfigured
-from django.db.models import DateTimeField, CharField, SlugField, Q, UniqueConstraint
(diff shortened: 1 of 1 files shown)


django_extensions/db/fields/__init__.py
```diff
@@ -45,6 +47,24 @@ def _get_fields(model_cls):
...
+        fields = []
+
+        if django.VERSION >= (4, 0) and unique_constraint.contains_expressions:
+
+            for expression in unique_constraint.expressions:
+                if isinstance(expression, F):
+                    fields.append(expression.name)
+                else:
+                    for inner_expression in expression.get_source_expressions():
+                        fields.append(inner_expression.name)
```
p-r-a-v-i-n, line 64: Does `get_source_expressions()` always return objects with a .name attribute? For example, I think Value or some Func subclasses may not. Are you sure this is safe?

p-r-a-v-i-n:
Would be great if you could add some failing test cases for these changes.

Issue #1493, opened 2020-02-24
Dumpscript serializes enumfields.Enum incorrectly


DanLipsitt:
Enum fields made with [django-enumfields](https://github.com/hzdg/django-enumfields) serialize incorrectly as `__repr__`-like strings.

### Model
```python
import enumfields

class Action(enumfields.Enum):
    LIST = 'list'

class Permission(models.model):
   action = enumfields.EnumField(Action, max_length=30)
```

### `manage.py dumpscript` output (incorrect)
```python
app_permission_1 = UserCategoryPermission()
app_permission_1.action = <Action.LIST: 'list'>
```

### yaml `manage.py dumpdata` output (correct)
```yaml
- model: app.permission
  pk: 1
  fields:
    action: list
```

blueyed:
related (code): https://github.com/django-extensions/django-extensions/pull/1746

PR #1746, opened 2022-07-15
fix(dumpscript): handle formatting of tuples

blueyed:
With tuples it would fail to format it:

```
  File "…/django_extensions/management/commands/dumpscript.py", line 360, in get_waiting_list
    value = get_attribute_value(self.instance, field, self.context, force=force, skip_autofield=skip_autofield)
  File "…/django_extensions/management/commands/dumpscript.py", line 697, in get_attribute_value
    item_locator = orm_item_locator(value)
  File "…/django_extensions/management/commands/dumpscript.py", line 77, in orm_item_locator
    clean_dict[key] = str("%s" % v)
TypeError: not all arguments converted during string formatting
```


blueyed:
With regard to https://github.com/django-extensions/django-extensions/issues/1493 it should maybe use `repr` instead?!

aysudeliahmetoglu (approved):

PR #1754, opened 2022-08-12
Added support for Django 4.1.

felixxm:

PR #1736, opened 2022-06-22
Use list values for requires_system_checks

danlamanna:
See the changes in Django 4.1 for https://docs.djangoproject.com/en/dev/releases/4.1/#features-removed-in-4-1.

PR #1758, opened 2022-08-23
Replace requires_system_checks False values with []

shangxiao:
Support for boolean values for `requires_system_checks` was [removed in Django 4.1](https://docs.djangoproject.com/en/4.1/releases/4.1/#features-removed-in-4-1) (and deprecated since 3.2 so safe to replace).

The replacement for `False` is `[]`: https://github.com/django/django/commit/c60524c658f197f645b638f9bcc553103bfe2630


shangxiao:
Ah just saw #1736 & #1754

PR #1759, opened 2022-08-24
Ensure consistent output in show_urls for django 4.0+

jeremy-engel:
Updates `show_urls` to properly show the view class for class-based
views in django 4.0 and later. There was a regression in django 4.0
that was fixed in https://github.com/django/django/commit/f4b06a3cc1e54888ff86f36a1f9a3ddf21292314
that resulted in `view` being output in `show_urls` instead of the
actual class name.

The tests are essentially returned to the state there were in before https://github.com/django-extensions/django-extensions/commit/c8d4c4acdd92090e7bc86a8e663431d8631de2a4.


trbs:
Thanks !

PR #1760, opened 2022-08-30
Add a py.typed file to export type annotations

brianhelba:
In order for any of the type annotations used within django-extensions to be available to downstream users, PEP-561 requires that an empty `py.typed` marker file be present as package data. For more details, see:
https://blog.whtsky.me/tech/2021/dont-forget-py.typed-for-your-typed-python-package/

This allows downstream users to have type checking when directly using any Python APIs which happen to have type annotations.

More importantly, it allows Django and DRF mixin classes (like `ModelUserFieldPermissionMixin`) and subclasses (like `TimeStampedModel`) to inherit type information if type stub packages for Django or DRF are in use.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ for dirpath, dirnames, filenames in os.walk(extensions_dir):
         path = os.path.join(*relative_path)
         package_files = package_data.setdefault('.'.join(parts), [])
         package_files.extend([os.path.join(path, f) for f in filenames])
+package_data['django_extensions'].append('py.typed')
 
 
 version = __import__('django_extensions').__version__


adamchainz:
I like to see Django projects rolling out type hints. Good work. That said, I would encourage improving the type hints in the package before exposing them to users.

I've learned from typing my own packages that packages should use Mypy's strict mode (which enables several flags) in order for users to also be able to use strict mode. If a package has some type hints but isn't fully covered, or conformant with the other flags in strict mode, it can make it *harder* for projects to adopt typing. Rather than the package being typed as `Any` throughout, a partially typed package requires users to use a lot of `# type: ignore` comments, or `cast()`s, to use it, even correctly.

When I add 

```
$ cat pyproject.toml
[tool.mypy]
strict = true
```

trbs:
It would be really good if we can make this as clean as possible before committing it into the code base...
We should also add a check for this in the pipeline I think, so we avoid regressions later...

trbs:
Yeah I meant having it error out on the mypy step of the pipeline :-)

sshishov:
Hello guys, I can see 2 years passed and no work is done. I completely understand that it is almost impossible to add typing to the whole repo in one go especially without conflicts etc.

Therefore there should be another approach here, you should add `py.typed` and start adding types to your repo with strict flag (file by file) or (package by package) or even (function by function).

Eventually we can come up with typed repository.

What do you think about it? Otherwise it will never be done... Especially if you want the help from the community!

brianhelba:
Looks like other PRs have now fully accomplished this (adding `py.typed` and exporting it as package data).

PR #1762, opened 2022-08-30
Add option to ignore files on runserver_plus reload

### LINKED ISSUE #1753: add exclude_patterns to werkzeug reloader
Hi,

I couldn't find any way to exclude certain files/dir from triggering server reloading.

We can specify extra files to watch via `RUNSERVER_PLUS_EXTRA_FILES = []`, but what about excluding?

Werkzeug does accept it via:
`werkzeug.serving.run_simple(exclude_patterns="")`
https://werkzeug.palletsprojects.com/en/2.2.x/serving/#werkzeug.serving.run_simple

Would be good to have it, thank you ❤️

marciomazza:
Fix #1753

In line with [Werkzeug's api](https://werkzeug.palletsprojects.com/en/2.2.x/serving/#werkzeug.serving.run_simple) added the option of ignoring file patterns for reload. 

On a similar way to the existing
`runserver_plus ... --extra-file` option and the  
`RUNSERVER_PLUS_EXTRA_FILES` setting,

added the 
`runserver_plus ... --exclude-pattern` option and the
`RUNSERVER_PLUS_EXCLUDE_PATTERNS` setting

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -165,6 +165,8 @@ class Command(BaseCommand):
                                  'Either --cert-file or --key-file must be provided to use SSL.')
         parser.add_argument('--extra-file', dest='extra_files', action="append", type=str, default=[],
                             help='auto-reload whenever the given file changes too (can be specified multiple times)')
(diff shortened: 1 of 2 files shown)


trbs:
Thanks !

PR #1769, opened 2022-10-29
Add REMOTE_USER to werkzeug environment

### LINKED ISSUE #1708: runserver_plus with REMOTE_USER ?
Just curious if runserver_plus supports the REMOTE_USER option, if so, I can't seem to find a way to make it work (and thusly could use a little guidance).

When running the default `python manage.py runserver` command I can get it to work with automatically logging in a REMOTE_USER by doing this: `REMOTE_USER="johndoe" python manage.py runserver 8000`

Then I can use the site locally as 'johndoe' already logged in.

However, when I replace `runserver` with `runserver_plus` in that example, it just returns me to the login page - I can't seem to find any configuration that will work with the REMOTE_USER setting and `runserver_plus`; am I missing something or does that feature not work with `runserver_plus` command?

confuzeus:
Django RemoteUserMiddleware and RemoteUserBackend are able to authenticate users by looking at the REMOTE_USER request header.

In Werkzeug, this header isn't set unlike in Django's default WSGI server.

This patch add the variable to the Werkzeug environ if it's set.

Fixes #1708

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -326,6 +326,9 @@ class Command(BaseCommand):
                 environ = super().make_environ()
                 if not options['keep_meta_shutdown_func'] and 'werkzeug.server.shutdown' in environ:
                     del environ['werkzeug.server.shutdown']
+                remote_user = os.getenv('REMOTE_USER')
+                if remote_user is not None:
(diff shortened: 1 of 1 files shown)


confuzeus:
Can't find a good way to test this. Suggestions welcome!

PR #1775, opened 2022-12-08
runserver_plus template reloading

### LINKED ISSUE #1766: [runserver_plus] autoreload django
Hello, 
after updating Django to version 4.1.1 I noticed that using runserver_plus on the modification of python files, the autoreload works correctly but on the modification of html files if I do not restart the server,  by reloading the web page the changes are not displayed. If I use the Django runserver command everything works fine.
I think this all depended on this: https://docs.djangoproject.com/en/4.1/releases/4.1/#templates

I wait your feedback.
Thanks!

foarsitter:
The [cached.Loader](https://docs.djangoproject.com/en/4.1/ref/templates/api/#django.template.loaders.cached.Loader) template loader is default in [Django 4.1](https://docs.djangoproject.com/en/4.1/releases/4.1/#templates), even with `DEBUG=True`. 

To take advantage of this loader in development the `runserver` command is extended to invalidated the cache when an [template is changed](https://github.com/django/django/blob/9da2210f121085e998dc7cedb783bc4568390b92/django/utils/autoreload.py#L369). 

`runserver_plus` does not listen to changes in `.html` and does not invalidate the template cache on change. 

This pull requests adds the `.html` suffix to `extra_files` so the [Werkzeug reloader](https://github.com/pallets/werkzeug/blob/3115aa6a6276939f5fd6efa46282e0256ff21f1a/src/werkzeug/_reloader.py#L271) fires the correct signals. By firing the `django.utils.autoreload.file_changed` event on each change the `django.template.autoreload.template_changed` receiver can trigger `reset_loaders` when needed. 

Fixes #1766

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -7,6 +7,7 @@ import sys
 import traceback
 import webbrowser
 import functools
+from pathlib import Path
 from typing import List, Set
 
 import django
@@ -15,7 +16,7 @@ from django.core.management.base import BaseCommand, CommandError, SystemCheckEr
 from django.core.management.color import color_style
 from django.core.servers.basehttp import get_internal_wsgi_application
 from django.dispatch import Signal
-from django.utils.autoreload import get_reloader
+from django.utils.autoreload import file_changed, get_reloader
 from django.views import debug as django_views_debug
 
 try:
@@ -86,11 +87,17 @@ if HAS_WERKZEUG:
 
             @property
             def extra_files(self):
-                return self._extra_files.union(_error_files)
+                return self._extra_files.union(_error_files, {"*.html"})
 
             @extra_files.setter
             def extra_files(self, extra_files):
                 self._extra_files = extra_files
+
+            def trigger_reload(self, filename: str) -> None:
+                path = Path(filename)
+                results = file_changed.send(sender=self, file_path=path)
+                if not any(res[1] for res in results):
+                    super().trigger_reload(filename)
 
         _reloader.reloader_loops[name] = WrappedReloaderLoop
 


foarsitter:
The actions fail due to #1777

foarsitter:
@trbs thanks for mergin. Is there something of a schedule for a new release? A new release would make it possible for https://github.com/cookiecutter/cookiecutter-django/pull/4028 to upgrade to Django 4.1.

foarsitter:
@bmihelac `extra_files` is added to the patterns over here: https://github.com/pallets/werkzeug/blob/main/src/werkzeug/_reloader.py#L328

Can you tell me more about your setup? How does your template loader settings look like? Do you use the StatReloader?

bmihelac:
@foarsitter thanks for your time.

I believe that template loaders are standard from what you get when you start new project + DIRS:

```python
 TEMPLATES = [
    {
        "BACKEND": "django.template.backends.django.DjangoTemplates",
        "DIRS": [
            BASE_DIR / "templates",
        ],
        "APP_DIRS": True,
        "OPTIONS": {
            "context_processors": [
                "django.template.context_processors.debug",
                "django.template.context_processors.request",
                "django.contrib.auth.context_processors.auth",
                "django.contrib.messages.context_processors.messages",
            ],
        },
    },
]
```

If `extra-file` argument is similar to what this patch is doing, here is the easiest way to reproduce this.

```bash
./manage.py runserver_plus --extra-file="*.html"
```

Changing `test.html` with this command would not trigger reloading.

```bash
./manage.py runserver_plus --extra-file="test.html" 
```

Changing `test.html` with this would reload and it is visibile in console.

```
 * Detected change in '.../test.html', reloading
 * Restarting with stat
```

let me know if you need any more information.

foarsitter:
The template reload does reset the cache of the cached.loader, it does not restart werkzeug. So there is no trace in the logs. 

But the evidence you provide is strong. 

I verified the patch and here it works like a charm. Im on Linux using Docker, using the statloader and Django 4.1 

How do you install the main branch through pip? What OS you are on?

bmihelac:
I am on linux, Django 4.1, stat reloader, installed django-extensions with:

```bash
pip install git+https://github.com/django-extensions/django-extensions
```

I have tried once more right now and cached template is displayed.

foarsitter:
Do you have time to craft a minimal viable project and share it on GitHub? Currently Im on holiday but next week I can dive into it.

bmihelac:
@foarsitter sure, i'll create test project for this ASAP

foarsitter:
@bmihelac you are right about that the `StatReloaderLoop` does not work. Somehow I was using `WatchdogReloaderLoop` which gives the desired results. 

`werkzeug._reloader._find_stat_paths` should be patched or we need to add all the template files to `extra_files`. What do you think is the desired solution?

bmihelac:
@foarsitter what do you think about using [dango.template.autoreload.get_template_directories](https://github.com/django/django/blob/main/django/template/autoreload.py#L10) to simplify getting template directories?

I wonder about the case of adding new templates after dev server is started - would they be watched or `extra_files` list is created only when server starts?

foarsitter:
@bmihelac my attempt to mimic `get_template_directories` is indeed a poor one. Thanks for pointing me to the existence of that method. 

`runserver`  detects the addition of new files, my proposal does not. So that is a problem.

foarsitter:
> If `extra-file` argument is similar to what this patch is doing, here is the easiest way to reproduce this.
> 
> ```shell
> ./manage.py runserver_plus --extra-file="*.html"
> ```

This brought my on an idea to create a workaround except each `--extra-file` is wrapped with  `os.path.abspath` so only `.html` in the root directory is watched.

danjesus:
I'm having the same problem here, changes to templates don't trigger a change event.

Only when using the default runserver.

Anyone else with same problem?

PR #1777, opened 2022-12-12
Autoupdate pre-commit hooks & fix linting for runserver_plus.py

### LINKED ISSUE #1776: flake8 pre-commit hooks version outdated
Since the version of flake8 is not pinned, the Github actions use a different version of flake8 then pre-commit. This gives inconsistent outcomes between  the linter and pre-commit action.

foarsitter:
This pull request fixes #1776 by updating the pre-commit hooks to there latest version. The change in runserver_plus.py:76 is needed to satisfy mypy.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -73,7 +73,7 @@ DEFAULT_POLLER_RELOADER_INTERVAL = getattr(settings, 'RUNSERVERPLUS_POLLER_RELOA
 DEFAULT_POLLER_RELOADER_TYPE = getattr(settings, 'RUNSERVERPLUS_POLLER_RELOADER_TYPE', 'auto')
 
 logger = logging.getLogger(__name__)
-_error_files: Set = set()
+_error_files: Set[str] = set()
 
 
 if HAS_WERKZEUG:


ulgens:
Flake8 has some breaking changes and bugs with the 6.0 version. I wouldn't recommend updating it to the latest for now.

foarsitter:
As state in the issue, we are already using `flake8` in the actions since the version is not specified. This pull request updates pre-commit to the latest version so the `Pre-Commit Checks` & `Linters` action both use flake8==6.0.0 and do not yield different results anymore. 

What problems do you encounter using flake8? I did not need to make any modifications to get the CI to pass.

ulgens (approved):

django_extensions/management/commands/runserver_plus.py
```diff
@@ -73,7 +73,7 @@
 DEFAULT_POLLER_RELOADER_TYPE = getattr(settings, 'RUNSERVERPLUS_POLLER_RELOADER_TYPE', 'auto')
 
 logger = logging.getLogger(__name__)
-_error_files = set()  # type: Set[str]
+_error_files: Set = set()
```
browniebroke, line 76: I'm curious as to why you dropped the `str`? Was it not correct? Like this, I mean:
```suggestion
_error_files: Set[str] = set()
```

foarsitter: Good catch, wasn't intended.

Issue #1778, opened 2022-12-12
Pipchecker fails on Python 3.11 with pip >= 22.1


foarsitter:
When running the tests with Python 3.11 and pip 22.1 or greater an AttributeError is thrown as shown below. It has something to do with https://pip.pypa.io/en/stable/news/#id64 

```
AttributeError: 'PathDistribution' object has no attribute 'project_name'
```

```
/PycharmProjects/django-extensions/venv4/lib/python3.11/site-packages/_distutils_hack/__init__.py:33: UserWarning: Setuptools is replacing distutils.
  warnings.warn("Setuptools is replacing distutils.")
Traceback (most recent call last):
  File "/PycharmProjects/django-extensions/manage.py", line 13, in <module>
    execute_from_command_line(sys.argv)
  File "/PycharmProjects/django-extensions/venv4/lib/python3.11/site-packages/django/core/management/__init__.py", line 446, in execute_from_command_line
    utility.execute()
  File "/PycharmProjects/django-extensions/venv4/lib/python3.11/site-packages/django/core/management/__init__.py", line 440, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/PycharmProjects/django-extensions/venv4/lib/python3.11/site-packages/django/core/management/base.py", line 402, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/PycharmProjects/django-extensions/venv4/lib/python3.11/site-packages/django/core/management/base.py", line 448, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/PycharmProjects/django-extensions/django_extensions/management/utils.py", line 62, in inner
    ret = func(self, *args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/PycharmProjects/django-extensions/django_extensions/management/commands/pipchecker.py", line 141, in handle
    self.check_pypi()
  File "/PycharmProjects/django-extensions/django_extensions/management/commands/pipchecker.py", line 167, in check_pypi
    name = dist.project_name
           ^^^^^^^^^^^^^^^^^
AttributeError: 'PathDistribution' object has no attribute 'project_name'
```

PR #1780, opened 2022-12-12
Add Python 3.11 support

foarsitter:
This pull requests adds Python 3.11 to Github actions and classifiers. It mitigates the issue with `importlib_metatdata` as described in #1778. 

Linting fails due to https://github.com/django-extensions/django-extensions/pull/1777


trbs:
With https://github.com/django-extensions/django-extensions/issues/1778 merged do we still need _PIP_USE_IMPORTLIB_METADATA ?

foarsitter:
Your are correct, it is no longer required.

PR #1786, opened 2022-12-27
Add Python 3.11 support

foarsitter:
By mistake I closed #1780 but here is the rebase with main and without `_PIP_USE_IMPORTLIB_METADATA`


trbs:
great thanks !

foarsitter:
Thank you for your quick response @trbs

PR #1794, opened 2023-01-27
Allow fileglobs in the dont_load list

ianepperson:
Permits file globbing definitions in the `dont_load` list by creating a list subclass called PathList which inspects file globs too. When this change is applied, a user can use `*`, `?` and even `[seq]` to fine tune the `dont_load` list.

This change _does not_ use regex, and thus does not effect existing functionality whatsoever - the glob characters are not permitted within a Python identifier.

```python
SHELL_PLUS_DONT_LOAD = [
    "my_app.Tagulous_*",  # Don't load any of the Tagulous models defined in my_app
    "*.Tagulous_*",  # Don't load any of the Tagulous models in any app
    "my_app.model[ABC]",  # Don't load modelA, modelB nor modelC
    "my_app.model?",  # Don't load any of the above, plus any other modelD, modelE
]
```
---
I'm using the Tagulous plugin which creates additional models every time I add a tag field. Each of these models is prefixed with `my_app.Tagulous_` and fails to load in shell_plus. I keep adding to the DONT_LOAD list, and I'd rather do that once.
```python
SHELL_PLUS_DONT_LOAD = [
    "my_app.Tagulous_*",
]
```


ianepperson:
Can I get a maintainer to approve running the workflows to ensure this works as advertised?

trbs:
Can you rebase this PR ?

PR #1796, opened 2023-01-28
runserver_plus autoreload on template change

foarsitter:
This pull-request is the second attempt (see #1775 for details) to get the correct behavior for reloading of the development server. 

The `watchdog` reloader did accept the *.html wildcard pattern but the `stat` reloader isn't. 

Therefore this pull-request adds an list of all files inside all template directories to `extra_files`. Even non .html files inside the template directories are triggering a reload since it can occur other extensions are used for templates like .css & .js. New files added to the template directory also trigger a reload since this is the same behavior as `runserver`. 

Thanks to @bmihelac for providing feedback to my first attempt.

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -16,6 +16,7 @@ from django.core.management.base import BaseCommand, CommandError, SystemCheckEr
 from django.core.management.color import color_style
 from django.core.servers.basehttp import get_internal_wsgi_application
 from django.dispatch import Signal
+from django.template.autoreload import get_template_directories
 from django.utils.autoreload import file_changed, get_reloader
 from django.views import debug as django_views_debug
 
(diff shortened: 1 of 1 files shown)


bmihelac:
I have tested this and can confirm that it reloads updated and new files.

nattyg93:
Hi folks, I've just run into this one myself is this likely to be merged any time soon?

wgordon17:
Just chiming in to agree that this has solved the issue for me as well (where it was _not_ solved for me in #1775). @trbs any hope of getting a review on this please? 🙏 

@foarsitter I suspect it would help if you address the failing pytest checks

foarsitter:
Thanks for testing @wgordon17. 

If I recall correctly, the tests are fixed in #1797 as reported in #1812.

trbs:
Guess the flip side is that it can be annoying as well that files which has nothing to do with "runserver" triggers a reload... 

Maybe this kind of behavior would also be nice to be controlled via some cli options && settings..

danjesus:
Hey guys, anyone else still having this problem?

Not working with `runserver_plus`, only with `runserver`.

With django-extensions version 3.2.3

foarsitter:
Open an issue with a minimal failing example if you think the problem still exists. It should be solved in 3.2.3.

PR #1798, opened 2023-02-05
Cleanup http: links

Crocmagnon:
Move http: links to https: where possible and update old links. Also removed broken links.

I was triggered by the old werkzeug link in `runserver_plus`.


trbs:
Awesome thanks !

PR #1800, opened 2023-02-09
Add django-prometheus DB backends support

mounirmesselmeni:
List of backends provided by django-prometheus:

https://github.com/korfuri/django-prometheus/tree/master/django_prometheus/db/backends

This will fix the issue of not being able to reset_db as the backend is unknown.

A workaround I was using in `settings.py` it the following:

```python
if 'reset_db' not in sys.argv:
    DATABASES['default']['ENGINE'] = DATABASES['default']['ENGINE'].replace(
        'django.db.backends', 'django_prometheus.db.backends'
    )
```


trbs:
Thanks !

PR #1808, opened 2023-03-24
Add django_prometheus support

pickfire:
Support django_prometheus database wrappers for reset db

https://github.com/korfuri/django-prometheus#monitoring-your-databases


pickfire:
Looks like a duplicate of https://github.com/django-extensions/django-extensions/pull/1800

Issue #1773, opened 2022-11-30
show_urls --decorator broken


takis:
Hi,

Using the `show_urls` management command with the `--decorator` option is broken.
```./manage.py show_urls```

In commit 702f798f47c18d5135709fea698307f4f32fb08e the parameters for `fmtr.format` were turned into separate variables. The newly created `decorator` variable overwrote the existing `decorator` variable which held a list of decorator names.

Pull request django-extensions/django-extensions/pull/1737 fixes this specific problem.

Unfortunately, even after applying the aforementioned patch, it still does not work as expected.

**Comment by takis** [comment]
With pull request django-extensions/django-extensions/pull/1737 applied, all views in a module in which a decorator was imported, get marked as being decorated, which is also not what I would expect.

```bash
./manage.py show_urls   -d login_required --format json
```


```json
[
    {
        "url": "/login/",
        "module": "django.contrib.auth.views.LoginView",
        "name": "login",
        "decorators": ""
    },
    {
        "url": "/logout/",
        "module": "django.contrib.auth.views.LogoutView",
        "name": "logout",
        "decorators": ""
    },
    {
        "url": "/admin/",
        "module": "django.contrib.admin.sites.index",
        "name": "admin:index",
        "decorators": ""
    },
    {
        "url": "/admin/login/",
        "module": "django.contrib.admin.sites.login",
        "name": "admin:login",
        "decorators": ""
    },
    {
        "url": "/admin/logout/",
        "module": "django.contrib.admin.sites.logout",
        "name": "admin:logout",
        "decorators": ""
    },
    {
        "url": "/admin/password_change/",
        "module": "django.contrib.admin.sites.password_change",
        "name": "admin:password_change",
        "decorators": ""
    },
    {
        "url": "/admin/password_change/done/",
        "module": "django.contrib.admin.sites.password_change_done",
        "name": "admin:password_change_done",
        "decorators": ""
    },
    {
        "url": "/admin/autocomplete/",
        "module": "django.contrib.admin.sites.autocomplete_view",
        "name": "admin:autocomplete",
        "decorators": ""
    },
    {
        "url": "/admin/jsi18n/",
        "module": "django.contrib.admin.sites.i18n_javascript",
        "name": "admin:jsi18n",
        "decorators": ""
    },
    {
        "url": "/admin/r/<int:content_type_id>/<path:object_id>/",
        "module": "django.contrib.contenttypes.views.shortcut",
        "name": "admin:view_on_site",
[...]

PR #1815, opened 2023-05-05
debug for decorator option

ryanchen99:


ryanchen99:
debug for decorator option in command "show_urls"

ryanchen99:
In the for loop of the original code, each iteration overrides the variable decorator, so the problem was that only one url with the specified decorator could be found. The issue is reported here: [issues/1773](https://github.com/django-extensions/django-extensions/issues/1773/) I changed the variable name so that the variable decorator was not affected, which solved the problem.

trbs:
Maybe I am missing something but I do not see how this changes anything...

line 122 might scope bleed the "decorator" variable but it's reassigned anyways on line 140....

Otherwise add a test-case for this so we can demonstrate this fixes it.

PR #1837, opened 2023-09-15
Call post_command when the command raises an unhandled exception

jrobichaud:
`post_command` is currently not called if an exception is raised which can cause problems if we really need it.


trbs:
Thanks !

Issue #1831, opened 2023-07-31
`mail_debug` incompatible with Python 3.12


michel-slm:
Python 3.12 dropped `asyncore` which has been deprecated since 3.6, recommending `asyncio` instead as the replacement:

https://docs.python.org/3.12/whatsnew/3.12.html#removed

weslord:
Also `smtpd`

bgtron:
It would be wonderful to get this resolved.  Just wanting to _bump_ this issue...

sundayz-hunter:
I've just updated to 4.0 released yesterday.
The error persists: it's no longer `asyncore `but `aiosmtpd` that's causing the problem.

`Error fetching command 'mail_debug': Please install 'aiosmtpd' library to use mail_debug command.`

browniebroke:
Correct, the command has been rewritten and now requires an 3rd party library to be installed. It's documented here: https://django-extensions.readthedocs.io/en/latest/mail_debug.html

Rationale as to why it's no installed by default: https://github.com/django-extensions/django-extensions/issues/1923

PR #1841, opened 2023-10-18
Update mail_debug command to use aiosmtpd

weslord:
Issue #1831

Python 3.12 dropped `asyncore` and `smtpd`

This adds a new dependency `aiosmtpd` from pypi. Should we add some kind of check and instructional message if the library is not found?

### DIFF
--- a/django_extensions/management/commands/mail_debug.py
+++ b/django_extensions/management/commands/mail_debug.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
-import asyncore
+import asyncio
 import sys
+from aiosmtpd.controller import Controller
 from logging import getLogger
-from smtpd import SMTPServer
 from typing import List
 
 from django.core.management.base import BaseCommand, CommandError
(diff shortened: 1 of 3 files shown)


foarsitter (approved):

foarsitter (approved):

jayvdb:
I think it would be useful to add additional tests - currently there is only one, which doesnt test that the server is doing smtp stuff.

And in the `Command.handle`, add something like https://github.com/django-extensions/django-extensions/blob/0e5747586332333f572bcfbf674b712e11aa69dd/django_extensions/management/commands/export_emails.py#L137-L141 to check that `aiosmtpd` is installed.

ulgens:
@weslord Any luck with the recommended changes? I can create a PR to help.

trbs:
We are not always consistent but often we do have a try-except with error message for the missing import.
Given that we are changing this behavior that would be better. 

Could you also add documentation for mail_debug (under docs/mail_debug.rst) ?
Then we can highlight the requirement there as well.

trbs:
Thanks, #1880 is merged 👍

PR #1842, opened 2023-10-20
Add support for Django 5.0

tim-schilling:
This allows the test suite to pass. However this can't be merged as is. The command `list_signals` should probably be updated to include the `is_async` descriptor in its output somewhere. You can see how the `Signal.receivers` list of tuples was modified to include `is_async` here: https://github.com/django/django/blob/stable/5.0.x/django/dispatch/dispatcher.py#L116

If you want to make the change for `list_signals` on your own, I won't be offended. If you'd like me to make the change, I will try to make more time for it in the future.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -26,6 +26,7 @@ jobs:
           - "40"
           - "41"
           - "42"
+          - "50"
           # GH Actions don't support something like allow-failure ?
           # - "master"
         exclude:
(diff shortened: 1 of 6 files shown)


browniebroke:
I've hit an issue with Django 5.1, noticed that 5.0 support was missing and ended up finding this PR. I've opened a PR targeting this branch: https://github.com/django-extensions/django-extensions/pull/1842

mgrdcm:
@tim-schilling do you think this can be merged now that @browniebroke's change has been integrated with your branch?

mgrdcm:
> @mgrdcm I don't think that was the blocker. The blocker was the `list_signals` question.

Ahh, I thought perhaps that was addressed with this line: https://github.com/tim-schilling/django-extensions/pull/1/files#diff-46a7fd41cc63527799c0c091f05529ae1bf281b919847edb5bca6e9537341f95R20

trbs:
Thanks !

PR #1848, opened 2023-11-18
graph_models style per app

ahmad88me:
# Idea
Style the generated model per app. 

# use case
 Imagine you're working on a Django project composed of two distinct applications, namely 'app1' and 'app2'. These applications are interconnected, with models in each app having relationships not only within the same app but also across the two apps. To effectively visualise these intricate relationships, you're aiming to generate a comprehensive graph model. This graph will not only illustrate the internal model structures within each individual app but will also clearly depict the inter-app model relationships. The goal is to have a single, unified graphical representation that vividly differentiates the models of 'app1' and 'app2', while simultaneously showcasing the connections between them. Such a visual tool would be invaluable for understanding the complex architecture of your Django project at a glance.

# How to use it
You can either have a `.app-style.json` or you can specify the style file using the `--app-style` option. It takes a path to the style file. The style file is a json file which has the app name as the key and the style as the value. Here is a sample one:

```
{
  "app1": {"bg": "#341b56"},
  "django.contrib.auth": {"bg": "#561b4c"},
  "app2":{"bg": "#1b3956"}
}
```

Further, I updated the django style to support the background change. So in the future, this can easily be extended to more styles (e.g., color, font, ..., etc.)


django_extensions/__init__.py
trbs, line 1: Could you please remove this from the PR ?

ahmad88me: Done

trbs:
Thanks for the update. It looks like the PR now removes `django_extensions/__init__.py` entirely ?

And sorry with the move to `pyproject.toml` and `ruff` `graph_models.py` changed..

ahmad88me:
@trbs no worries 😉  .. I added the `__init__` and resolved the conflicts.

ahmad88me:
@trbs resolved the new conflicts and fix the lint

ulgens:
@ahmad88me Is it possible to update https://django-extensions.readthedocs.io/en/latest/graph_models.html#graph-models too? Currently, there is no info in the docs about this new option.

trbs:
> @ahmad88me Is it possible to update https://django-extensions.readthedocs.io/en/latest/graph_models.html#graph-models too? Currently, there is no info in the docs about this new option.

good catch, yes please submit a new PR for updated docs, thanks !

Issue #1835, opened 2023-08-24
CommandError: Could not load shell runner: 'IPython Notebook'.


flo-85:
I am using django 4.2 abd python 3.10.
I am using Windows WSL (Linux subsystem on a windows machine)
Jupyter notebook is installed.

This works:
```
./manage.py shell_plus --ipython
````

But not this:
```
./manage.py shell_plus --notebook
```

Error:
```
Traceback (most recent call last):
  File "/home/florent/github/web_api/.venv/lib/python3.10/site-packages/django_extensions/management/commands/shell_plus.py", line 281, in get_notebook
    from notebook.notebookapp import NotebookApp
ModuleNotFoundError: No module named 'notebook.notebookapp'

CommandError: Could not load shell runner: 'IPython Notebook'.
```

pip list:
```
ipython                           8.14.0
ipython-genutils                  0.2.0
jupyter                           1.0.0
jupyter_client                    8.3.0
jupyter-console                   6.6.3
jupyter_core                      5.3.1
jupyter-events                    0.7.0
jupyter-lsp                       2.2.0
jupyter_server                    2.7.2
jupyter_server_terminals          0.4.4
jupyterlab                        4.0.5
jupyterlab-pygments               0.2.2
jupyterlab_server                 2.24.0
jupyterlab-widgets
```

**Comment by slyapustin** [comment]
I was able to temporarily fix that by using old version of `jupyter_client`:
> jupyter_client==7.4.8

**Comment by aa403** [comment]
also getting this issue, and it continues to present after downgrading jupyer_client ... I have the following:

```
ipython==8.15.0
ipython-genutils==0.2.0
  
jupyter==1.0.0
jupyter-console==6.6.3
jupyter-events==0.7.0
jupyter-lsp==2.2.0
jupyter_client==7.4.8  # downgraded from 8.3.1
jupyter_core==5.3.1
jupyter_server==2.7.3
jupyter_server_terminals==0.4.4
jupyterlab==4.0.5
jupyterlab-pygments==0.2.2
jupyterlab-widgets==3.0.8
jupyterlab_server==2.24.0


notebook==7.0.3  # it looks like the issue is here??
notebook_shim==0.2.3

```

aa403:
yes, `--lab` does work ...  but its quite a jump -- this django-extensions should be updated for it, no?

clax70rm:
I solved the issue downgrading notebook with this command

pip install notebook==6.5.6

Then using the command 

[...]

PR #1852, opened 2023-12-16
Add small mention to --lab flag

soof-golan:
# What

Mention an already existing `--lab` flag in the shell_plus docs

# Why

It's a useful feature and easy to miss if one is not actively looking for it


sidmitra:
This seems useful to add to the docs. I didn't know about this until i saw https://github.com/django-extensions/django-extensions/issues/1835

trbs:
Thanks !

PR #1857, opened 2024-01-16
Add webp support for graph_models

kviktor:
Adds `webp` support, graphviz already seem to support it https://graphviz.org/docs/outputs/webp/.

Example images generated: https://szi.app/shared/erd.png vs https://szi.app/shared/erd.webp

```
     12K jan   16 22.38 erd.png
    7,1K jan   16 22.40 erd.webp
```


ulgens (approved):

trbs:
Thanks !

PR #1863, opened 2024-03-09
Use python_history for shell_plus command

igorgaming:
This feature [was added to Django](https://github.com/django/django/commit/3921b1c6d24c9d5a60e5f5f83c9a394104089c21) a few years ago. It allows you to restore the command history when you re-enter the `shell` (in our case, `shell_plus`),

What it means:
If the user has executed some commands, they are written to `~/.python_history` by default.
Then, when exiting and re-entering, the entire history can be restored in a shell using the up arrow.

I just copied this code from the Django command, but if you want to do the same in other way - np with that.


igorgaming:
See #1869

PR #1869, opened 2024-04-07
Fix start up order and add history

anitahammer:
#1863

Comparing vs django/master:

Screenshot from 2024-04-08 02-20-31

FYI @igorgaming


trbs:
Thanks !

PR #1873, opened 2024-05-10
Deterministic output for .dot

XF-FW:
Hey. My workflow involves generating a dot file on every commit and diffing with the current one, to see if there are new changes.
The current dot output has two things that may differ even with the same src:

- Timestamp
- Abstract model order

Added an option for `deterministic` to be able to set it as a constant value and fixed the shuffling done with the abstract models, via `list(set(x))`

### DIFF
--- a/django_extensions/management/modelviz.py
+++ b/django_extensions/management/modelviz.py
@@ -8,7 +8,8 @@ Based on:
   Adapted to be used with django-extensions
 """
 
-import datetime
+from collections import OrderedDict
+from datetime import datetime
 import os
 import re
 
@@ -105,6 +106,7 @@ class ModelGraph:
         self.rankdir = kwargs.get("rankdir")
         self.display_field_choices = kwargs.get("display_field_choices", False)
         self.ordering = kwargs.get("ordering")
+        self.deterministic = kwargs.get("deterministic", False)
 
     def generate_graph_data(self):
         self.process_apps()
@@ -121,9 +123,10 @@ class ModelGraph:
                             relation['needs_node'] = False
 
     def get_graph_data(self, as_json=False):
-        now = datetime.datetime.now()
+        date = datetime.fromtimestamp(0) if self.deterministic else datetime.now()
+
         graph_data = {
-            'created_at': now.strftime("%Y-%m-%d %H:%M"),
+            'created_at': date.strftime("%Y-%m-%d %H:%M"),
             'cli_options': self.cli_options,
             'disable_fields': self.disable_fields,
             'disable_abstract_fields': self.disable_abstract_fields,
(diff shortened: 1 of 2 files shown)


django_extensions/management/modelviz.py
```diff
@@ -119,9 +121,12 @@ def generate_graph_data(self):
                             relation['needs_node'] = False
 
     def get_graph_data(self, as_json=False):
-        now = datetime.datetime.now()
+        if self.deterministic:
+            date = datetime.datetime(2000, 1, 1, 0, 0)
```
trbs, line 125: Would it make more sense to use Epoch Time (`January 1, 1970`) here ?

`date = datetime.datetime.fromtimestamp(0)`

XF-FW: Agreed. Updated accordingly.

XF-FW:
I had written this test:

```
# tests/management/test_modelviz.py (lines 20-31)
def test_generate_graph_data_deterministic(self):
    app_labels = ["auth"]
    should_all_be_the_same = []

    for _ in range(10):
        graph_data = generate_graph_data(app_labels)
        dot = generate_dot(
            graph_data, "django_extensions/graph_models/original/digraph.dot"
        )
        should_all_be_the_same.append(dot)

    self.assertEqual(should_all_be_the_same[1:], should_all_be_the_same[:-1])
```

but it's not actually failing in main.

It has been a while since I've picked this up, but this continues to be a problem to me, so I would like for this feature to be supported. I think this PR should be at least a good first step for that.

django_extensions/management/modelviz.py
```diff
@@ -121,9 +123,10 @@ def generate_graph_data(self):
                             relation['needs_node'] = False
 
     def get_graph_data(self, as_json=False):
-        now = datetime.datetime.now()
+        date = datetime.fromtimestamp(0) if self.deterministic else datetime.now()
```
trbs, line 126: Can we change the variable name of `date` to `create_at` here ?

(When I am looking at the `date.strftime` below, I find myself questioning if this is the Python builtin date class aka `from datetime import date` or something else... So I think that would make it far more clear to read the code)

trbs:
> I had written this test:
> 
> ```
> # tests/management/test_modelviz.py (lines 20-31)
> def test_generate_graph_data_deterministic(self):
>     app_labels = ["auth"]
>     should_all_be_the_same = []
> 
>     for _ in range(10):
>         graph_data = generate_graph_data(app_labels)
>         dot = generate_dot(
>             graph_data, "django_extensions/graph_models/original/digraph.dot"
>         )
>         should_all_be_the_same.append(dot)
> 
>     self.assertEqual(should_all_be_the_same[1:], should_all_be_the_same[:-1])
> ```
> 
> but it's not actually failing in main.
> 
> It has been a while since I've picked this up, but this continues to be a problem to me, so I would like for this feature to be supported. I think this PR should be at least a good first step for that.

Happy to try and get this merged 👍 

I am sorry but with the `ruff format` changes a lot of files have changes syntax, including these.. which is probably what is creating the merge conflicts now.

Would you have some time to rebase this PR and get those fixed ?

Also adding a test would be really really great.

jrd:
There is also a problem when using `--display-field-choices` as the current code is doing:
```python
    if self.display_field_choices and field.choices is not None:
        choices = {c for c, _ in field.choices}
        t = str(choices)
```
The choices order will be in different order with different calls. A probable fix would be `t = str(sorted(choices))`.

I can take over the rebase, add test and this specific case if @XF-FW doesn’t show up…

PR #1878, opened 2024-06-19
Add setting to fix `runserver_plus` debugger on custom host names

browniebroke:
Add `RUNSERVERPLUS_TRUSTED_HOSTS` setting to work with Werkzeug 3.0.3+ debugger.

[Werkzeug version 3.0.3](https://werkzeug.palletsprojects.com/en/3.0.x/changes/#version-3-0-3) introduced a fix for a security vulnerability:

> Only allow `localhost`, `.localhost`, `127.0.0.1`, or the specified hostname when running the dev server, to make debugger requests. Additional hosts can be added by using the debugger middleware directly. The debugger UI makes requests using the full URL rather than only the path. [GHSA-2g68-c3qc-8985](https://github.com/advisories/2g68-c3qc-8985)

I'm running `runserver_plus` with Werkzeug 3.0.3 on a machine with a custom dev host (e.g. `dev.mydomain.com`) which means that the Werkzeug debugger no longer works.

As far as I could tell, there was no way to tell Werkzeug to trust my dev host, so adding this option to do so.

## Questions

1. Should this default to `settings.ALLOWED_HOSTS`? It might be a good enough default on a Django project...
2. Updated the docs, let me know if I missed anything?

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -454,6 +454,11 @@ class Command(BaseCommand):
             if self.nopin:
                 os.environ['WERKZEUG_DEBUG_PIN'] = 'off'
             handler = DebuggedApplication(handler, True)
+            # Set trusted_hosts (for Werkzeug 3.0.3+)
+            try:
+                handler.trusted_hosts = settings.RUNSERVERPLUS_TRUSTED_HOSTS
+            except AttributeError:
+                pass
 
         runserver_plus_started.send(sender=self)
         run_simple(
--- a/docs/runserver_plus.rst
+++ b/docs/runserver_plus.rst
@@ -85,10 +85,18 @@ An ajax based console appears in the pane and you can start debugging.
 Notice in the screenshot above I did a `print environ` to see what was in the
 environment parameter coming into the function.
 
-*WARNING*: This should *never* be used in any kind of production environment.
-Not even for a quick problem check.  I cannot emphasize this enough. The
-interactive debugger allows you to evaluate python code right against the
-server.  You've been warned.
+.. warning::
+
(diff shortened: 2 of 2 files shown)


browniebroke:
Looks like the test failures are unrelated to my change, or am I missing something? 

Is there any interest for this change?

jribbens:
> Is there any interest for this change?

Yes, this change, or something like it, is vital for `runserver_plus` to continue being useful.
(Pretty disgraceful of Werkzeug to make such a change in a patch release frankly.)

jribbens:
Given this project appears to be moribund, the following work-around can be used to avoid having to pin Werkzeug to <3.0.3.

Put the following in your `settings/dev.py` file or whatever you have that's similar:
```python
import werkzeug.debug

werkzeug.debug.DebuggedApplication.check_host_trust = (
    lambda *args, **kwargs: True  # noqa: ARG005
)

File: django_extensions/management/commands/runserver_plus.py [hunk 1]
```diff
@@ -454,6 +454,11 @@ def make_environ(self):
             if self.nopin:
                 os.environ['WERKZEUG_DEBUG_PIN'] = 'off'
             handler = DebuggedApplication(handler, True)
+            # Set trusted_hosts (for Werkzeug 3.0.3+)
+            try:
+                handler.trusted_hosts = settings.RUNSERVERPLUS_TRUSTED_HOSTS
```
trbs, line 459: Can't we do something like:
```
trusted_hosts = getattr(settings, 'RUNSERVERPLUS_TRUSTED_HOSTS', None)
if trusted_hosts:
    handler.trusted_hosts = trusted_hosts 
```

That way we avoid avoid the try-except block and are still defensive in not settings it unless the setting actually exists.

PR #1880, opened 2024-07-15
Update mail_debug command to use aiosmtpd

ulgens:
This PR copies the commit from https://github.com/django-extensions/django-extensions/pull/1841 and adds an "Is aiosmptd installed?" check, similar to the recommendation on https://github.com/django-extensions/django-extensions/pull/1841#issuecomment-1928494777 .

I'm open to recommendations about the wording of the error message and potential tests.


trbs:
Thanks !

PR #1860, opened 2024-02-06
Fixes #1652

wadewilliams:
Ran into this issue myself, figured I would PR it instead of adding a comment. 

Thanks to @danaspiegel for opening the issue

PR #1883, opened 2024-08-26
Update sqldiff.py to support geospatial databases

julopezc-acciona:
The code later tries to access geo_col.name and geo_col.type_code, so we add those attributes.
I think it's trying to access geo properties for tables that do not have them, as the database has some geo tables


julopezc-acciona:
Closed in favor of https://github.com/django-extensions/django-extensions/pull/1860 which seems a better solution

Issue #1836, opened 2023-09-09
runjob or runjobs will fail on an App if there happens to be a file named "jobs.py" in the root dir of that app


jiansongyang:
When performing command `python3 manage.py runjob -l` or `python3 manage.py runjobs -l` on an App where a file named `jobs.py` is in the root direcotory of that app (example could be django-rq):
```
Traceback (most recent call last):
  File "manage.py", line 16, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 413, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 354, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 398, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.6/site-packages/django_extensions/management/utils.py", line 62, in inner
    ret = func(self, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django_extensions/management/commands/runjob.py", line 56, in handle
    print_jobs(only_scheduled=False, show_when=True, show_appname=True)
  File "/usr/local/lib/python3.6/site-packages/django_extensions/management/jobs.py", line 151, in print_jobs
    jobmap = get_jobs(when, only_scheduled=only_scheduled)
  File "/usr/local/lib/python3.6/site-packages/django_extensions/management/jobs.py", line 120, in get_jobs
    path = find_job_module(app_name, subdir)
  File "/usr/local/lib/python3.6/site-packages/django_extensions/management/jobs.py", line 83, in find_job_module
    return module.__path__[0]
AttributeError: module 'django_rq.jobs' has no attribute '__path__'
```
The error roots from management/jobs.py when walking through all apps to find out their jobs:
```
def find_job_module(app_name: str, when: Optional[str] = None) -> str:
    """Find the directory path to a job module."""
    parts = app_name.split('.')
    parts.append('jobs')
    if when:
        parts.append(when)
    module_name = ".".join(parts)
    module = importlib.import_module(module_name)
    return module.__path__[0]

```
[...]

PR #1887, opened 2024-09-16
Handle non-package modules when looking for job definitions

ghickman:
This fixes an error with runjobs when a jobs.py exists in the environment.  Modules have no `__path__` member defined unless it's a package, so we should be able to rely on that here.

Fix: #1836


trbs:
Thanks !

PR #1890, opened 2024-11-04
Updates output text to include SVG format

bsdtux:
While looking through the cli help for graph_models it was not clear if I could use SVG for the output of a file. This PR will hopefully help anyone that is wondering if they can use SVG as an output instead of just PNG or JPG


bsdtux:
Looks like Ruff reformatted the whole file as well. Going to revert and try again

Issue #1829, opened 2023-07-05
Add svg to the documentation of the --output arg of graph_models


mardukbp:
In the current version of django-extensions (3.2.3) the documentation of the command `graph_models`

```bash
python manage.py graph_models --help
```

only mentions png and jpg as possible outputs. Searching in the GH issues I learned that svg is also supported. This should be included in the documentation. SVG is useful since it is not currently possible to specify the size or the dpi of the bitmap output  (see #1429).

varunsaral:
I think it will be easy to do. Just have to add the in svg in 
https://github.com/django-extensions/django-extensions/blob/dd794f1b239d657f62d40f2c3178200978328ed7/django_extensions/management/commands/graph_models.py#L95-L99

PR #1891, opened 2024-11-05
Updates graph models command with output extension type formats

bsdtux:
Recently looking for a way to export our models to an ERD and discovered this project. However wasn't sure from the help text if SVG was available. Hoping that this PR will allow / inform anyone using the graph_models command that SVG is a format available


bsdtux:
I didn't realize until now that this PR would address this issue https://github.com/django-extensions/django-extensions/issues/1829

PR #1898, opened 2024-12-06
Improve error message for missing shell_plus import

stianjensen:
Today, if you misspell something in SHELL_PLUS_IMPORTS, you just get a line printed when opening the shell, with a list of names in scope from the module you have an incorrect import from. This can be a bit confusing to debug.

This commit adds some more context to this print, both to explain what's happening, and tell you which module import is at fault.


trbs:
Thanks !

PR #1900, opened 2024-12-09
Fix generate_password command on Django 5.1

### LINKED ISSUE #1897: ./manage.py generate_password raises an AttributeError: 'BaseUserManager' object has no attribute 'make_random_password'
The `generate_password` command stopped working in django 5.1. It raises an error now:
```
...
  File "proj/venv/lib/python3.10/site-packages/django_extensions/management/commands/generate_password.py", line 30, in handle
    return manager.make_random_password()
AttributeError: 'BaseUserManager' object has no attribute 'make_random_password'
```

browniebroke:
Fix #1897 

The `make_random_password` method has been removed in Django 5.1. Looking at the most recent implementation, it was calling `get_random_string` under the hood, and that seems to exists in all versions of Django supported by django-extensions

### DIFF
--- a/django_extensions/management/commands/generate_password.py
+++ b/django_extensions/management/commands/generate_password.py
@@ -1,13 +1,9 @@
 # -*- coding: utf-8 -*-
 from typing import List
 
+from django.core.management.base import BaseCommand
 from django.utils.crypto import get_random_string
 
-try:
-    from django.contrib.auth.base_user import BaseUserManager
-except ImportError:
-    from django.contrib.auth.models import BaseUserManager
-from django.core.management.base import BaseCommand
 from django_extensions.management.utils import signalcommand
 
 
--- a/django_extensions/management/commands/generate_password.py
+++ b/django_extensions/management/commands/generate_password.py
@@ -1,6 +1,8 @@
 # -*- coding: utf-8 -*-
 from typing import List
 
+from django.utils.crypto import get_random_string
+
 try:
     from django.contrib.auth.base_user import BaseUserManager
 except ImportError:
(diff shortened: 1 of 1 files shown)


deronnax:
Any idea of why the CI is failing? All jobs are hitting GitHub's timeout. I would say a test in `test_managestate.py` is doing shady things, network related I guess.

The test setup could use some simplification, a tox that calls a makefile that calls pytest is too much, the Makefile could be gone, wdyt?

browniebroke:
I noticed the problem but didn't spent any time investigating... I noticed that you opened a PR to look at it, but haven't yet seen the results of the run (they seem to be queue'd?)

deronnax:
I need to workflow approval from a maintainer. If you are one, please approve :) 
Capture d’écran 2024-12-31 à 17 05 37

browniebroke:
I'm not but you should be able to change the workflow yaml to run on your own fork, on pushes

deronnax:
yep, that's what I did. I reproduce the problem on my fork. By using pytest-timeout, I will be able to know which test gets stuck. I will come back with a fix or a suggestion.

deronnax:
regarding the PR itself, are we sure we want to reintroduce something django removed ?

browniebroke:
> regarding the PR itself, are we sure we want to reintroduce something django removed ?
> 
> EDIT: original rational in https://code.djangoproject.com/ticket/33758
> 
> > The usage of
> > User.objects.make_random_password()
> > is limited since the generated passwords are not valid because they were created without meeting the validators criteria.

That's actually a great point, thanks for bringing this up. Removing it would make sense to me, yes.

Issue #1899, opened 2024-12-08
AutoSlugField is ASCII only


amirreza8002:
related issue: #1809 

https://github.com/django-extensions/django-extensions/blob/main/django_extensions/db/fields/__init__.py#L231-L234

as you can see in the above code, the `allow_unicode` parameter of [slugify](https://github.com/django/django/blob/main/django/utils/text.py#L452) is not passed, so it only works as an ASCII utility

is there any chance we get a fix?

browniebroke:
You can pass a slugify function to the field, so that should work:

```python
def slugify_unicode(value):
    return slugify(value, allow_unicode=True)

class MyModel(models.Model):
    title = models.CharField(max_length=42)
    slug = AutoSlugField(
        populate_from="title", 
        slugify_function=slugify_unicode,
    )
```

amirreza8002:
hmm
should've thought that myself 🫠

but I do hope for django extension to add support for it at some point 

thanks 🙏

mscheper:
> You can pass your own slugify function to the field, so that should work

@browniebroke: Thanks for explaining this reasonable workaround.

However, since `allow_unicode` is the only additional parameter for Django's stock `slugify` function, it does seem intuitive for it to be passed in the constructor. I admit I didn't look closely enough at the API, but it certainly was what I expected, and it seems @amirreza8002 also expected it, and perhaps the author of #1809.

It would save devs from writing quite a boilerplate function, and I think a lot of them would. It made sense for `allow_unicode` to default to `False` in the dim dark years of IE, but I don't think there's much reason to prohibit unicode in URLs in the '20s, especially given the inconvenience for users unfamiliar with the Latin alphabet.

So I suggest this issue be reopened, and interpreted as a feature request. While I realise an `allow_unicode` argument would be ignored if either the model or the constructor defined a custom slug function, I would argue that this is still more intuitive than the argument not being passed.

PR #1901, opened 2024-12-09
Split field extensions page into sections

browniebroke:
Someone recently asked [a question](https://github.com/django-extensions/django-extensions/issues/1899) about `AutoSlugField` and I found the page difficult to grok:

- It wasn't clear at first glance that there was multiple fields documented
- Once I saw that, it took me a moment to see where the AutoSlugField ended

Splitting each field into its own section will solve both these problems. Maybe it might have avoided the question in the first place.

### DIFF
--- a/docs/field_extensions.rst
+++ b/docs/field_extensions.rst
@@ -3,30 +3,31 @@ Field Extensions
 
 :synopsis: Field Extensions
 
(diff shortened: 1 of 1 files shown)


trbs:
Thanks !

Issue #1892, opened 2024-11-12
Moving django-extensions forward


browniebroke:
The project development [seems to have slowed down](https://github.com/django-extensions/django-extensions/graphs/contributors), new issues and pull request are unanswered, so at this point it's showing all the signs of being unmaintained. I love some of the tools in this project, and would miss them if they stopped working. I think it's one of the main Django 3rd party packages, so I would hate to see it turn stale.

I'm not throwing any shade at the current maintainers, maintaing a project of such magnitude over time is no small feat and it's still working pretty well despite the lack of love in the past few months. However, by providing so many integrations with other libraries beyond Django (Jupyter notebook, 3.13 REPL, Werkzeug, watchdog, ..), if it's not kept up to date, the ecosystem keeps moving and cracks start to appear...

So I'd like to start a discussion to see how can we move this project forward. Do the current maintainers still have the energy and interest to pick this up (could be a temporary hiatus)? Does the project need new ones to be added? Would it help to be under a new organisation like @django-commons?

I hope this doesn't come across negatively.

_PS: I've opened a discussion in the google group earlier, but it didn't show up, I assume new messages are hidden until moderated._

trbs:
It's true that work on Django Extensions (from me) has been slowed down to
a crawl in recent times. Leading to PR's remaining open and it slowly
drifting from meanline Django and other dependencies/related projects.

The reasons for this are mostly typical for open source projects, many
articles and blogs [1] have been written by other projects and maintainers
and a lot of that applies for Django Extensions as well.

I definitely don't consider the project to be abandoned and therefore this
should be a temporary hiatus, however I cannot tell exactly when work will
pick up in a significant way.

The easiest and quickest way would be when an entity would spend some
actual money on the project allowing me (and ideally even others) to work
on it during our working hours. While none have any (other than moral)
obligation to do so, I know that projects like Django Extensions (and there
are many others too) are valuable to me when working on corporate or client
projects so that does not seem like that big of an ask.

I want to make crystal clear that I cannot understate how much I appreciate
all the people that donated to me or the project and companies like Sentry
that recently donated money (
https://blog.sentry.io/we-just-gave-750-000-dollars-to-open-source-maintainers/
) to keep the project going and to allow me to spend time on it. Though in
all earnest those never amounted to anything more than tens of dollars, the
biggest one being github's one time $500 donation, not close to anything
enough to pay the bills.

I hope this doesn't come across negatively.

No worries :-)

[...]

PR #1909, opened 2025-01-17
Fix broken tests to get CI passing

browniebroke:
Pin test dependencies to make as much tests pass as possible.


tests/test_dumpscript.py
```diff
@@ -78,22 +79,21 @@ def test_with_datetimefield(self):
...
         # 'date_joined': dateutil.parser.parse("2019-05-20T03:32:27.144586+09:00")
-        with open(dumpscript_path + '/test.py', 'wt') as test:
+        with (dumpscript_path / 'test.py').open('wt') as test:
             call_command('dumpscript', 'django_extensions', stdout=test)
 
         # Check dumpscript without exception
         call_command('runscript', 'test')
 
-        # Delete dumpscript
-        shutil.rmtree(dumpscript_path)
```
browniebroke, line 96: I noticed that this didn't run and left some files on disk if one of the `call_command` above crashed.

tests/test_dumpscript.py
```diff
@@ -78,22 +79,21 @@ def test_with_datetimefield(self):
...
 
-        dumpscript_path = './django_extensions/scripts'
+        dumpscript_path = Path(__file__).parent.parent / 'django_extensions' / 'scripts'
+        dumpscript_path.mkdir(parents=True, exist_ok=True)
+        (dumpscript_path / '__init__.py').touch()
 
-        os.mkdir(dumpscript_path)
-        open(dumpscript_path + '/__init__.py', 'w').close()  # for python 2.7
+        # Delete dumpscript at the end of the test
+        self.addCleanup(shutil.rmtree, dumpscript_path)
```
browniebroke, line 87: I learned about `enterContext` from [this Adam Johnson's post](https://adamj.eu/tech/2024/12/30/python-temporary-files-directories-unittest/) and seemed like it would help here.

As the folder needs to be in a specific location, the solution from the post wasn't exactly what I needed, but that helped me find this `addCleanup` method which is used under the hood by `enterContext`.

requirements-dev.txt
```diff
@@ -2,11 +2,12 @@ shortuuid
 werkzeug
 django_pdb
 python-dateutil
+pytest<8
 pytest-django
 pytest-cov
 factory-boy
 requests
-pygments
+pygments<2.18
```
browniebroke, line 10: Something changed between Pygment 2.17 and 2.19 (installed on Python 3.8) which changes `HighlightTagTests.test_should_highlight_python_syntax_with_name`. 

It seems that Pygment 2.18+ doesn't support Python 3.7, so we get a different output in the test depending on the Python version...

As Python 3.7 is EOL, next step would be to drop support for it, unpin this version and update the test to match the newer output.

browniebroke:
Ok we're getting there... The last test failing are due to `asyncore` being dropped in Python 3.12 as reported in https://github.com/django-extensions/django-extensions/issues/1831

Looks like there is already a work in progress to resolve that: https://github.com/django-extensions/django-extensions/pull/1841 so I might leave that aside (for now at least)

requirements-dev.txt
browniebroke: > As Python 3.7 is EOL, next step would be to drop support for it, unpin this version and update the test to match the newer output.

Next step here: https://github.com/django-extensions/django-extensions/pull/1910 (includes some changes from this PR)

browniebroke:
@trbs In case you missed this... Ideally we need to upgrade to pytest 8 but I went for an easy fix for now by pinning it to <8

browniebroke:
> Is there anything I can help with, as an external contributor? 

I don't know to be honest, hopefully the project development will get picked up at some point and activity will resume (https://github.com/django-extensions/django-extensions/issues/1892)

> I feel like this will enable a lot of other changes.

Same 💯

ulgens (approved):

ulgens:
@browniebroke Did something happen on the main branch? This PR looked useful.

browniebroke:
Fixes have been ported to the main branch:

- https://github.com/django-extensions/django-extensions/commit/58340065927d7c76bb5c9787e5b592c3624a5345
- https://github.com/django-extensions/django-extensions/commit/2d1f1e61dda3d075616f3c50b2fc6283122ec6ab
- https://github.com/django-extensions/django-extensions/commit/b79504ae2c5a9033d7f56e3f8d9e1fc9197a94da
- https://github.com/django-extensions/django-extensions/commit/86d029ddeaf92505a4852de24f5c7f2058906f3d

To the point where the diff was pretty much empty

PR #1912, opened 2025-01-19
Add readthedocs config file and requirements

browniebroke:
Looks like the docs have stopped building a bit more than a year ago: https://app.readthedocs.org/projects/django-extensions/

I assume the webhooks have stopped sending and might need an admin to reconnect GitHub and RTD. 

While looking at that, I noticed that there is no config file for RTD. At time of writing, this is now deprecated and is going to be required from tomorrow: https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

I've setup a test project from my fork and it's building ok from my branch: https://app.readthedocs.org/projects/browniebroke-django-extensions/builds/26898547/

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,22 @@
+# Read the Docs configuration file for Sphinx projects
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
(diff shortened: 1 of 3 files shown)


trbs:
Thanks !

PR #1927, opened 2025-04-14
Verify named URLs

robert-kisteleki:
This is a working version of a feature that aims to answer the question: "is this named URL defined somewhere?" The use case is when one wants to remove views and needs to check if those are still referred from existing templates *somehere* in the project. Failing to remove references leads to template rendering errors (i.e. 500s).

It combines a bit of code from `validate_templates` and `show_urls` - mostly because the functionality combines the two approaches. It'd be nice to remove code duplication but that needs restructuring of existing code, in particular making `extract_views_from_urlpatterns()` available as a function elsewhere.

It can be easily tested by referring to some non-existing name (in the form of `{% url 'xxx' %}`) in a template and running the command.

Limitation: as it's written it only works with Django templates (ie. not Jinja).

The command is useful as it is; ideas for refinements (or a merge) are welcome!


trbs:
Think we can move some common code like `extract_views_from_urlpatterns` to `django_extensions.utils`.

Some quick feedback;
- Missing docs 
- Code seems like it could be cleaner (things such as the `add_name` function)
- Wondering if the name is intuitive... maybe it's me... but looking at the name I was not really clear what `verify_named_urls` would do... but I could not think of anything that would be better... so maybe this is not a real issue 😆

robert-kisteleki:
> Think we can move some common code like `extract_views_from_urlpatterns` to `django_extensions.utils`.

That'd help the cause indeed. Do you want me to make that part of this PR?

> Some quick feedback;
> 
>     * Missing docs

No problem, I can add it, just wanted first to have running code and see if the idea sticks

>     * Code seems like it could be cleaner (things such as the `add_name` function)

I'll take a look

>     * Wondering if the name is intuitive... maybe it's me... but looking at the name I was not really clear what `verify_named_urls` would do... but I could not think of anything that would be better... so maybe this is not a real issue 😆

I find it descriptive and as short as I could make it :-) but it can be changed if needed ofc

django_extensions/management/commands/verify_named_urls.py
```diff
@@ -0,0 +1,235 @@
...
+                except ImportError:
+                    continue
+                if namespace and p.namespace:
+                    _namespace = "{0}:{1}".format(namespace, p.namespace)
+                else:
+                    _namespace = p.namespace or namespace
+                pattern = describe_pattern(p)
+                if isinstance(p, LocaleRegexURLResolver):
+                    for language in self.LANGUAGES:
+                        with translation.override(language[0]):
```
trbs, line 200: `translation` is undefined ? (see github actions)

trbs:
Safety is not working anymore.... if you rebase it should be gone and the check should not fail anymore.
There some other failures as well that the linter is catching...

Also could you add a documentation for the command as well ?
Thanks !

django_extensions/management/commands/verify_named_urls.py
robert-kisteleki: Yeah, that's in a function that should be moved out (as discussed above). I added the relevant import into this file to fix this for now.

robert-kisteleki:
> Safety is not working anymore.... if you rebase it should be gone and the check should not fail anymore. There some other failures as well that the linter is catching...
> 
> Also could you add a documentation for the command as well ? Thanks !

I believe I fixed the linter hints, rebased and added a doc file.

PR #1930, opened 2025-04-23
Update list of default fields for list_model_info command

TomasLoow:
Add a few methods that have been added in newer Django versions to the `default_methods` list for the `list_model_info` command. This prevents them from being listed under every model in the project.

Method names added to the list:

- adelete
- arefresh_from_db
- asave
- get_constraints
- validate_constraints


trbs:
Thanks !

PR #1933, opened 2025-05-06
Feature: add show_template_paths management command

benbacardi:
This PR adds a management command `show_template_paths` that lists the full, absolute file system paths that are searched by the template loader when loading templates, in the order they are searched.

This is useful when debugging "template not found" errors, used in conjunction with the existing `find_template` management command.

```bash
$ python manage.py show_template_paths
/Users/joe.smith/my_project/templates
/Users/joe.smith/Library/Caches/pypoetry/virtualenvs/test_venv-py3.13/lib/python3.13/site-packages/django/contrib/admin/templates
/Users/joe.smith/Library/Caches/pypoetry/virtualenvs/test_venv-py3.13/lib/python3.13/site-packages/django/contrib/auth/templates
/Users/joe.smith/Library/Caches/pypoetry/virtualenvs/test_venv-py3.13/lib/python3.13/site-packages/django_extensions/templates
/Users/joe.smith/my_project/my_app/templates
```


django_extensions/management/commands/show_template_paths.py
```diff
@@ -0,0 +1,24 @@
...
+from django.template.engine import Engine
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = "List template search paths."
+
+    def handle(self, *args, **options):
+
+        test_template_name = str(datetime.datetime.now().timestamp())
```
trbs, line 13: Any specific reason to use the timestamp here ?

Instead of something like
```
test_template_name = "this_template_name_does_not_exist"
```

trbs: While true, both are highly unlikely... and the benefit of a fixed string is that it would be deterministic...

tests/management/commands/test_show_template_paths.py
```diff
@@ -0,0 +1,38 @@
...
+from pathlib import Path
+
+from django.core.management import call_command
+from django.test import TestCase
+
+
+BASE_DIR = Path(__file__).resolve().parent.parent.parent.parent
+
+
+class ShowPermissionsTests(TestCase):
```
trbs, line 12: please fix the name :-)

trbs:
Could you please fix the issues with the tests?

PR #1935, opened 2025-05-07
explicitly exclude __pycache__ dirs from runserver file change detection

navxio:
initialised `RUNSERVER_PLUS_EXCLUDE_PATTERNS` in settings.py


trbs:
Thanks !

PR #1946, opened 2025-07-25
standardise runserver_plus settings

saz:
There's a mixture of runserver_plus settings:
RUNSERVER_PLUS_...
RUNSERVERPLUS_...

This PR makes it possible to specify both variants in all cases, where currently `RUNSERVER_PLUS_...` is set. It shouldn't be backwards incompatible

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -85,10 +85,14 @@ ansi_escape = re.compile(
 )
 DEFAULT_PORT = "8000"
 DEFAULT_POLLER_RELOADER_INTERVAL = getattr(
-    settings, "RUNSERVERPLUS_POLLER_RELOADER_INTERVAL", 1
+    settings,
+    "RUNSERVER_PLUS_POLLER_RELOADER_INTERVAL",
+    getattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_INTERVAL", 1),
 )
 DEFAULT_POLLER_RELOADER_TYPE = getattr(
-    settings, "RUNSERVERPLUS_POLLER_RELOADER_TYPE", "auto"
+    settings,
+    "RUNSERVER_PLUS_POLLER_RELOADER_TYPE",
+    getattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_TYPE", "auto"),
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


trbs:
Think this is a good idea.

I am wondering if we should not pull out all the places where we use settings variables and place them all at the top.

Because:
1. It's currently kinda messy dealing with settings all over the place, specially if we are also checking multiple versions
2. It's not obvious from the code if we prefer one style over the other. I think we really want to go with `RUNSERVER_PLUS` as all the environment variables normally match the command name. This is further evidenced in this patch by allowing both `RUNSERVER_PLUS_POLLER_RELOADER_INTERVAL` and `RUNSERVERPLUS_POLLER_RELOADER_INTERVAL` but not allowing `RUNSERVERPLUS_SHOW_MESSAGES` since the original name was already `RUNSERVER_PLUS_SHOW_MESSAGES`.

If we move everything to the top, we can do something like:

```
DEFAULT_POLLER_RELOADER_INTERVAL = getattr(settings, "RUNSERVER_PLUS_POLLER_RELOADER_INTERVAL", 1)
DEFAULT_POLLER_RELOADER_TYPE = getattr(settings, "RUNSERVER_PLUS_POLLER_RELOADER_TYPE", "auto")
RUNSERVER_PLUS_SHOW_MESSAGES = getattr(settings, "RUNSERVER_PLUS_SHOW_MESSAGES")
...

# backwards compatibility
if not hasattr(settings, "RUNSERVER_PLUS_POLLER_RELOADER_INTERVAL") and hasattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_INTERVAL"):
    DEFAULT_POLLER_RELOADER_INTERVAL = getattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_INTERVAL")
if not hasattr(settings, "RUNSERVER_PLUS_POLLER_RELOADER_TYPE") and hasattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_TYPE"):
    DEFAULT_POLLER_RELOADER_TYPE = getattr(settings, "RUNSERVERPLUS_POLLER_RELOADER_TYPE")
...
```

I am open for a cleaner solution, but at least this keeps the style we want people to use upfront and center, including the default value and only falls back on the backwards compatibility value if it's set while the correct settings variable is not set.

saz:
I've taken the file name and https://github.com/django-extensions/django-extensions/blob/main/django_extensions%2Fmanagement%2Fcommands%2Frunserver_plus.py#L511 as hint, that the prefix is supposed to be `RUNSERVER_PLUS`.

Besides that, I've intentionally added only in those cases, where the prefix wasn't containing an underscore, because I think it would be the best to mark the old setting as deprecated (maybe in the future) and get rid of it, instead of introducing both ways and not having one prefix.

trbs:
Thanks !

I've merged your PR.

Hopefully I have some time soon, to do what I described above, so we can also deprecate the other style :-)

PR #1948, opened 2025-07-29
Fix import_items to match Python import behavior

Azd325:
I noticed after I migrated a project to the new style import for `SHELL_PLUS_POST_IMPORTS` that some imports didn't work anymore.

e.g.

```diff
SHELL_PLUS_POST_IMPORTS = [
-    "http.client",
+    "import http.client",
]

```

This change updates the import logic in import_items so that when importing a dotted module (like import http.client), only the top-level module (http) is added to the namespace, matching standard Python behavior. Submodules are still accessible as attributes (e.g., http.client). This makes shell imports more predictable and consistent with Python’s import system.


trbs:
Thanks !

PR #1966, opened 2025-11-24
Added `graph_app_model_dependencies` command

Ou7law007:
The new command `graph_app_model_dependencies` generates an app-level dependency graph based on Django model relations.

#### Builds a directed graph where:

- Nodes = Django app labels
- Edges = A → B if any model in app A has a relation (FK, O2O, M2M) to a model in app B

#### Supports the same output modes as graph_models:

- DOT (stdout or file)
- JSON
- pygraphviz / pydot image rendering (PNG, SVG, etc.)

#### Supports the same option patterns:

- --all-applications
- --output
- --layout, --rankdir, --ordering
- --json, --dot, --pygraphviz, --pydot

### Why

This provides a clear overview of inter-app coupling, which is helpful for:

- understanding migration ordering
- refactoring
- identifying apps' model-level cycles

### Example output

example_app_deps

Another example where apps' models are designed properly to have no cycles:
example_app_deps_no_cycles

### DIFF
--- a/django_extensions/management/commands/graph_app_model_dependencies.py
+++ b/django_extensions/management/commands/graph_app_model_dependencies.py
@@ -1,0 +1,423 @@
+from __future__ import annotations
+
+import json
+import os
+import sys
+import tempfile
+
+from django.apps import apps
+from django.core.management.base import BaseCommand
+from django.core.management.base import CommandError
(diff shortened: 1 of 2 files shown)


trbs:
thanks !

Issue #1886, opened 2024-09-10
KeyError: 'WERKZEUG_SERVER_FD' when running runserver_plus with noreload flag


Automatik:
I encountered an issue when trying to run my Django server using `runserver_plus` with `noreload` flag from `django-extensions` in a project created via `cookiecutter-django`.

The command I used is:
`python manage.py runserver_plus 0.0.0.0:8000 --noreload`

This raises the following error: ` KeyError: 'WERKZEUG_SERVER_FD'`

Full traceback
```
Traceback (most recent call last):
  File "/home/path/to/my/project/manage.py", line 31, in <module>
    execute_from_command_line(sys.argv)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django/core/management/base.py", line 412, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django/core/management/base.py", line 458, in execute
    output = self.handle(*args, **options)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django_extensions/management/utils.py", line 62, in inner
    ret = func(self, *args, **kwargs)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django_extensions/management/commands/runserver_plus.py", line 338, in handle
    self.inner_run(options)
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/django_extensions/management/commands/runserver_plus.py", line 459, in inner_run
    run_simple(
  File "/home/user/.venv/myvenv/lib/python3.10/site-packages/werkzeug/serving.py", line 1075, in run_simple
    fd = int(os.environ["WERKZEUG_SERVER_FD"])
  File "/usr/lib/python3.10/os.py", line 680, in __getitem__
    raise KeyError(key) from None
KeyError: 'WERKZEUG_SERVER_FD'

```

And these are my requirements:
base.txt
```
python-slugify==8.0.1  # https://github.com/un33k/python-slugify
Pillow==10.2.0  # https://github.com/python-pillow/Pillow
[...]

PR #1976, opened 2026-03-08
Fix KeyError on runserver_plus --noreload with Werkzeug 3.x

GitTinos:
Refs #1886

**Setup**
- Python 3.13
- Django 5.2.12
- Werkzeug 3.1.6

**Problem**
- `runserver_plus --noreload` forced `WERKZEUG_RUN_MAIN="true"`, but Werkzeug uses that flag to identify the real reloader child through [`is_running_from_reloader()`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/serving.py#L951-L957). Because of that, [`run_simple()`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/serving.py#L1088-L1092) tried to read `WERKZEUG_SERVER_FD` in no-reload mode, where no reloader parent had created it yet, causing `KeyError: 'WERKZEUG_SERVER_FD'`.
- Debugger setup was also split between `runserver_plus` and Werkzeug, which made ownership unclear and led to inconsistent behavior across reloader and no-reload paths.

**Solution**
- In `inner_run()`, `runserver_plus` no longer forces `WERKZEUG_RUN_MAIN`; it clears it in `--noreload` mode so Werkzeug follows its normal no-reload path.
- Debugger wrapping is now handled explicitly by `runserver_plus` in `wrap_handler_with_debugger()`, and the final Werkzeug call uses `use_debugger=False` to avoid double wrapping. `use_debugger=False` does not disable the debugger: the application is already wrapped with `DebuggedApplication`, so this only prevents Werkzeug from creating a second debugger layer inside [`run_simple()`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/serving.py#L1080-L1086).
- The debugger is applied only in processes that actually serve requests: the single-process no-reload case and the true reloader child, matching Werkzeug’s reloader flow in [`run_with_reloader()`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/_reloader.py#L436-L465).
- A follow-up adjustment keeps startup messages visible in `--noreload` and avoids duplicate debugger/PIN logs in reloader mode, where Werkzeug already logs them from [`DebuggedApplication`](https://github.com/pallets/werkzeug/blob/main/src/werkzeug/debug/__init__.py#L293-L301).

### DIFF
--- a/django_extensions/management/commands/runserver_plus.py
+++ b/django_extensions/management/commands/runserver_plus.py
@@ -393,7 +393,9 @@ class Command(BaseCommand):
         addrport = options["addrport"]
         startup_messages = options["startup_messages"]
         if startup_messages == "reload":
-            self.show_startup_messages = os.environ.get("RUNSERVER_PLUS_SHOW_MESSAGES")
+            self.show_startup_messages = (
+                not options["use_reloader"]
+            ) or os.environ.get("RUNSERVER_PLUS_SHOW_MESSAGES")
         elif startup_messages == "once":
             self.show_startup_messages = not os.environ.get(
                 "RUNSERVER_PLUS_SHOW_MESSAGES"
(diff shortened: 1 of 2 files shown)


GitTinos:
Hello @trbs , I’m a new contributor. While developing in my app I noticed this bug and looked for similar issues. I saw that some people are experiencing the same problem, but I understand that it may not be a priority. I tried to fix it. How does the process work? I noticed there are PRs that have been open for a while, but the codebase seems fairly up to date. How can I tell whether it has been reviewed?

P.S. I know everyone contributes in their free time; I would just like to understand how to receive feedback.

Thanks 💟

PR #1977, opened 2026-03-08
Fix some tests for Django 6

SuperSandro2000:
There are still two tests that fail but it is a start.


SuperSandro2000:
see #1979

PR #1993, opened 2026-08-13
fix: replace bare except with except Exception in dumpscript save_or_locate

TrueFurina:
Replace the bare `except:` in `save_or_locate()` with `except Exception:` — same intended behavior (print the object that failed to save), but KeyboardInterrupt/SystemExit are no longer intercepted first. Verified: `python -m py_compile` passes.


TrueFurina:
Closing due to inactivity - happy to revive if there is still interest. Thanks!

PR #1996, opened 2026-08-26
Added `graph_app_import_dependencies` command

Ou7law007:
Added a new management command that graphs app-level dependencies based on Python imports rather than model relations, and can detect and highlight import cycles between apps.

Existing python tools like pyan3 can do something similar to this but only for normal python modules. I thought this feature would be nice to have on a django-app level to give an easy way to audit apps and how they depend on each other as whole.

Another way to describe it is:
```
edges = { (app_of(file), app_of(imported_name))
          for app in django_apps
          for file in app.py_files
          for imported_name in ast_imports(file) }   # target resolved via Django's global app registry
```

### Why this is useful

`graph_models` and the existing `graph_app_model_dependencies` both graph relationships that show up as Django model fields (FK/M2M/O2O). Neither catches coupling that never shows up as a model relation — one app calling another's service layer, importing its Celery tasks, importing its admin classes, etc. This command finds that by parsing every installed app's source with ast and recording every edge where a file in one app imports something owned by another.

It also detects import cycles: non-trivial strongly connected components of the import graph (Tarjan's algorithm), plus the specific feedback edges that close each cycle.

### Example

app_import_dependencies_demo

```
# catalog/services.py — no deps

# inventory/services.py
import catalog.services

# billing/services.py
import catalog.services
import shipping.services          # <- cycle group 1

# shipping/services.py
import billing.services           # <- cycle group 1

# notifications/services.py
import billing.services
import shipping.services
import inventory.services

# reporting/services.py
import catalog.services
import billing.services
import inventory.services
import notifications.services
```
### Options

**Selection**
- `app_label` (positional, `nargs="*"`): app labels to include (and edges between them only)
- `--all-applications` / `-a`:  include every installed app instead

**Filtering**
- `--exclude`:  comma-separated app labels to drop
- `--exclude-external`:  keep only apps that live in your project (checked against the interpreter's real site-packages dirs, so it works regardless of where the venv itself lives, and correctly treats `pip install -e .` installs as local)
- `--exclude-name-prefix`:  comma-separated dotted app-name prefixes to drop, subtree included (e.g. `myproject.reporting.exporters` drops every nested app label under it, not just one label)
- `--exclude-file-pattern`:  comma-separated fnmatch globs for files to skip when scanning (e.g. `test*.py` to drop test-only imports like fixtures/factories, which are a different kind of coupling than production code; `migrations/*.py` to drop generated code). Nothing is excluded by default, and no test-file convention is hardcoded.

**Cycle detection**
- `--highlight-cycles`:  described above

**Output format** (mirrors `graph_models`/`graph_app_model_dependencies`)
- `--pygraphviz`, `--pydot`, `--dot`, `--json`:  pick the output format (default: DOT to stdout if none given and no `--output`)

**Rendering** (mirrors `graph_models`/`graph_app_model_dependencies`)
- `--output` / `-o`:  output file (extension picks format when no format flag is given)
- `--layout` / `-l`:  GraphViz layout engine (default `dot`)
- `--rankdir`:  `TB`/`BT`/`LR`/`RL` (default `TB`)
- `--ordering`:  `in`/`out` edge ordering

TODO:
- Add style customization especially for the cycle detection e.g. using command line args.


Ou7law007:
I'm confused by these last 3 CI errors! Can someone confirm they're a result of my changes?

PR #1979, opened 2026-03-10
Fix Django 6 support regressions (#1978)

asherzod1:
## Summary
This PR addresses the Django 6 regressions reported in #1978.

### What was fixed
1. **`list_signals` compatibility with Django 6 receiver tuples**
   - Django 6 uses a 4-item receiver tuple.
   - Updated unpacking logic for Django 6 while keeping compatibility with earlier versions.

2. **`RandomCharField` duplicate generation edge case**
   - `pre_save` can be invoked multiple times during insert in Django 6.
   - Reuse existing generated value when present to avoid unnecessary regeneration and `StopIteration` in duplicate test flow.
   - Made uniqueness collision check robust for mocked queryset behavior in tests.

3. **Constraint API update in test models**
   - Replaced deprecated `CheckConstraint(check=...)` usage with `condition=...` in test model definitions.

4. **Field class assertion compatibility**
   - Updated test assertion to accept `AutoField` or `BigAutoField` for `id` field output in `list_model_info` tests.

## Validation
Ran full test suite under Django 6:
- `pytest -q`
- Result: **533 passed, 4 skipped, 4 xfailed, 6 xpassed**

Closes #1978

PR #1997, opened 2026-08-31
Don't crash print-sql when last_executed_query returns None

### LINKED ISSUE #1956: SQL logger in shell plus crash with django-auditlog
Hi,

Some times, (not identified why for now) audit log create circumstances which make the django-extension code to crash.
It happen in `debug_cursor.py` in class `PrintQueryWrapperMixin` on line `raw_sql = self.db.ops.last_executed_query(self.cursor, sql, params)`
The function `last_executed_query` can return `None` (cf [db/backends/postgresql/operations.py file in django's code](https://github.com/django/django/blob/82b3b84a78055844ee07d5d97843a4fc72872e28/django/db/backends/postgresql/operations.py#L300)) 
In my case it occur because `cursor._query` is `None`.

The django-extension's problem is that a `None` value in raw_sql is not handled in PrintQueryWrapperMixin.execute. So it make crash later in `sqlparse.Lexer.get_tokens`.

Edit: The problem seems to occur on models with field of type `django.contrib.postgres.fields.DateTimeRangeField`

gyanu2507:
`PrintQueryWrapperMixin` always handed `last_executed_query()` to sqlparse. On PostgreSQL that can be `None` (DateTimeRangeField / django-auditlog), and the lexer then dies.

Fall back to the statement we just ran.

Fixes #1956

### DIFF
--- a/django_extensions/management/debug_cursor.py
+++ b/django_extensions/management/debug_cursor.py
@@ -69,6 +69,8 @@ def monkey_patch_cursordebugwrapper(
                 finally:
                     execution_time = time.time() - starttime
                     raw_sql = self.db.ops.last_executed_query(self.cursor, sql, params)
+                    if raw_sql is None:
+                        raw_sql = sql or ""
                     if truncate:
                         raw_sql = raw_sql[:truncate]
 
--- a/tests/management/commands/shell_plus_tests/test_shell_plus.py
+++ b/tests/management/commands/shell_plus_tests/test_shell_plus.py
@@ -1,5 +1,6 @@
 import os
 import re
+from unittest import mock
 import pytest
 import inspect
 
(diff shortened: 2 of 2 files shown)


gyanu2507:
Fixed my share of the lint failures in 6856e49 — the import and the context manager in the new test both ran past 88 columns, so `ruff-format` wanted them wrapped.

The remaining `ruff`/`precommit`/`mypy` failures are pre-existing on `main` and outside this diff:

- `ruff` E501 x3 and 2 files needing reformat: `django_extensions/db/fields/__init__.py` and `tests/testapp/models.py`
- `mypy`: `django_extensions/collision_resolvers.py:48`

`ruff` and `precommit` first went red at 32ddfb0 (`fix: add Django 6 compatibility ...`, #1979), which is this branch's parent — `aa59c37` and earlier are green. `mypy` has been failing further back than that; it's red on the last four commits to `main`. I reproduced both against a clean `origin/main` checkout: 3 ruff errors and 2 files to reformat, with no involvement from this branch.

So this PR took it from 3 files needing reformat down to the 2 that were already there. Happy to fix those in a separate PR if it'd help.

`tests/management/commands/shell_plus_tests/test_shell_plus.py` is 11 passed, and all pytest jobs were already green.

